Issue #2, opened 2015-03-24
Create  jsonschema and check the enums with it


rom1504:
See http://json-schema.org/ and https://www.npmjs.com/package/jsonschema

The point of this is to check automatically the json format documentation (see https://github.com/PrismarineJS/minecraft-data/wiki/Recipe-format) is actually true and can be trusted, even if recipes.json is updated or if the format change.

It would be nice to add this test to a ci service (circle ci ?)

PR #4, opened 2015-03-24
Convert old to new recipe format

Gjum:


rom1504:
I'm not sure it's a good idea to mix python and js in that project but maybe it's ok.

recipes_new.json should be recipes.json and it should be merged when we can handle it in mineflayer I think (that shouldn't be too hard to do, but I have been working on #2 until now)

Gjum:
You're right, mixing Python with Node might make the scripts difficult to use.
Also the conversion is a one-time thing, and the script is not needed again after that.
So I will remove it from the PR and rename `recipes_new.json` to `recipes.json`.

rom1504:
Ok your recipes.json respect your format at https://github.com/PrismarineJS/minecraft-data/wiki/Proposed-new-recipe-format .
I've checked with the new json format for recipes.

PR #31, opened 2015-08-13
Use countType where possible

roblabla:
#28

Issue #28, opened 2015-08-13
Use countType in array to reduce field count


roblabla:
Example : 

```json
{
  "name": "chunkDataLength",
  "type": "count",
  "typeArgs": {
    "type": "varint",
    "countFor": "chunkData"
  }
},
{
  "name": "chunkData",
  "type": "buffer",
  "typeArgs": {
    "count": "chunkDataLength"
  }
}
```

should become
```json
{
  "name": "chunkData",
  "type": "buffer",
  "typeArgs": {
    "countType": "varint"
  }
}
```

This greatly simplifies the protocol.json file, and makes it more obvious that count fields should not be supplied on write. I went over the 1.8 protocol, there are no instances where the count field is far away from the data it counts.

rom1504:
Sounds good, doing the .json transformation itself should be fairly easy, I think I'll do it and put it in a minecraft-data branch so we can see how to implement this in node-minecraft-protocol.

rom1504:
done

PR #32, opened 2015-08-13
Use countType where possible

roblabla:
#28 #31


rom1504:
need to fix https://github.com/roblabla/minecraft-data/blob/feature-countType/enums/protocol.json#L1884

roblabla:
Fixed. I think this is ready to be merged.

Issue #25, opened 2015-08-06
Put javascript stuff in tools/js


rom1504:
In order to be even more language independent, let's put js stuff in tools/js : 
* package.json
* bin/
* test.js
* removing index.js (along with that issue https://github.com/PrismarineJS/node-minecraft-data/issues/4 )

rom1504:
done

PR #34, opened 2015-08-17
Move js files, start fixing #25

rom1504:
To do : 
* [x] figure out what to do what history.md (and versions)


rom1504:
Almost done, fix this in the same time as https://github.com/PrismarineJS/node-minecraft-data/issues/4

PR #38, opened 2015-08-22
Add new type syntax [typeName, typeArgs]

roblabla:
Should be fairly straightforward. Not let's hope it's actually easier to use...

Also, this is missing the schema changes.


roblabla:
Looks ready

PR #41, opened 2015-08-23
Windows

### LINKED ISSUE #37: define the schema of a future Windows.json
* https://github.com/andrewrk/mineflayer/blob/master/lib/windows.js#L22
* https://github.com/SpockBotMC/SpockBot/blob/master/spock/plugins/helpers/inventory.py#L204

22:09 <+Gjum> it's just the window name (for protocol), readable name (although that can get set by server), named slot ranges, names for the properties, and later buttons
22:10 <+Gjum> properties are for example the smelting progress of a furnace
22:11 <+Gjum> buttons would just be list[id -> "name"]
22:12 <+Gjum> and then there are some edge cases like the horse, and the player itself
22:12 <+Gjum> horse has an entity id, player has armor and crafting, but no nameID

Gjum:
Adds a data file for the inventory windows.

Closes #37.

### DIFF
--- a/enums/windows.json
+++ b/enums/windows.json
@@ -1,0 +1,43 @@
+[
+  {
+    "id": "minecraft:beacon",
+    "name": "Beacon",
+    "slots": [
+      {
+        "name": "input",
+        "index": 0
+      }
+    ],
+    "properties": [
+      "level",
+      "effect one",
+      "effect two"
+    ],
+    "buttons": [
+      "apply",
+      "cancel"
+    ]
+  },
+  {
+    "id": "minecraft:brewing_stand",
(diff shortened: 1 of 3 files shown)


Gjum:
Maybe add

    "openedBy": [
      {type:"entity", id:100}
    ]

to specify how a window can be opened.

`type` can be `block` or `entity` at the moment.

rom1504:
Need to add a line in the README saying we add the data to that file manually.

Or maybe we can generate it using http://wiki.vg/Inventory#Windows (and possibly http://minecraft.gamepedia.com/Inventory#External_inventories )

PR #43, opened 2015-08-24
Introduce switch and void, remove condition

roblabla:
Switch is much more flexible, condition felt clunky and out of place. 

the idea of switch is to mimic C's switch/case statement.

- compareTo: The name of a field to be switched against.
- fields: An object where the key is a possible value of field, and the value is the type to use in case the field matches the key.
- default: Optional, if present, the type of what to read when field matched none of the values. If not present, the impl should error out as it means you have reached data outside the expected realm (values are expected to be exhaustive when default is not present).

void is also introduced. It is a type that does not read anything and returns null. It doesn't write anything and has a size of 0.

### DIFF
--- a/enums/protocol.json
+++ b/enums/protocol.json
@@ -454,45 +454,39 @@
                 {
                   "name": "velocityX",
                   "type": [
-                    "condition",
+                    "switch",
                     {
-                      "type": "short",
-                      "field": "intField",
-                      "values": [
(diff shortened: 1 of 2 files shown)


rom1504:
transformation https://gist.github.com/roblabla/c1cdb842fd6a9e1f9978 (useful for 1.9)

Issue #44, opened 2015-08-25
Complete the data of windows.json


rom1504:
see https://github.com/PrismarineJS/minecraft-data/issues/37

and https://github.com/PrismarineJS/minecraft-data/pull/41

rom1504:
Once the windows file is more complete, add a line in the extractors table in the readme

Gjum:
**Missing data:**
- [ ] anvil, horse, and villager slots
- [ ] most `openedWith` info
- [ ] beacon, enchanting, villager trading buttons (what happens when these are clicked?)
- [ ] accepted item(type)s for each slot (range) (think: crafting result, enchanting with lapis)

maybe also:
- [ ] put all(?) data on [wiki.vg](http://wiki.vg/Inventory)
- [ ] write an extractor script to get the data (and changes) from somewhere (e.g. mcwiki, wiki.vg, mc source)

Gjum:
I think we don't really need the buttons of a window.

- villager seems to use buttons just for looks client-side
- enchanting has its own packet
- beacon - no idea

rom1504:
Well maybe that means you can't automatically use the button data, but that doesn't mean that it isn't useful.

Gjum:
Its main use would be what packet you can send to simulate a button press. This however should not be modeled by a simple string.

The button-related info is also not very uniform. Something like in `protocol.json` would be needed, but I'm not sure if I like this option.

If you have an idea how to properly format that data, I'm ok with putting it in.

Gjum:
Sidenote: we could also look at other servers and see how they handle windows.
For example Cuberite:
http://api-docs.cuberite.org/cWindow.html and
http://api-docs.cuberite.org/cInventory.html

Gjum:
@Fenhl added `openedWith` info to http://wiki.vg/index.php?title=Inventory&diff=6832&oldid=6828

rom1504:
ah indeed, nice

rom1504:
windows.json added in readme

Gjum:
Since 1.9, the persistent 3*9 slots and the hotbar are no longer at the end.
This means that the schema needs to be changed to always include the position of the persistent slots and hotbar.

Gjum:
> ~~always include the position of the persistent slots and hotbar.~~

We could make that implicit: if it's not given, calculate maximum slot index and put inventory+hotbar after it; if it's given, use those indices.

I will implement it this way for SpockBot and see how it works out,

[...]

PR #47, opened 2015-08-25
Add all window types, remove buttons field

Gjum:
Adds all window types and removes the `buttons` field.

Related to #44.

### DIFF
--- a/enums/windows.json
+++ b/enums/windows.json
@@ -1,5 +1,25 @@
 [
   {
+    "id": "",
+    "name": "Player",
+    "slots": [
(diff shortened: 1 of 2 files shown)


rom1504:
I don't know why I didn't merge this, it's quite fine.

Gjum:
You didnt because there was some controversy regarding the buttons field.

rom1504:
Yeah but you're probably right, http://wiki.vg/Inventory#Windows doesn't have them and all.
If we really need them we can always add them back anyway.

PR #54, opened 2015-08-30
[WIP] Table-based layout for github.io/minecraft-data

roblabla:
Trying to match the functionality of wiki.vg/Protocol. Missing features :

- [x] Expanding switch
- [ ] Buffer countType


rom1504:
Nice

rom1504:
Something like this is enough to make the table take less space http://stackoverflow.com/questions/10687306/why-do-twitter-bootstrap-tables-always-have-100-width

rom1504:
Probably still need buffer countType, but good enough, nice, merging.

Issue #50, opened 2015-08-29
Improve the docgen of protocol values


rom1504:
http://prismarinejs.github.io/minecraft-data/#protocolTableA is useful but it would be even more useful if it displayed stuff in tables like wiki.vg

inspiration from the transformation in https://gist.github.com/rom1504/0d41cc284de99dfc8b6e

roblabla:
http://roblabla.github.io/minecraft-data/

rom1504:
Open a PR when you're done ;) 
There might be a few things to adjust relative to the changes I made in #53

rom1504:
* [ ] buffer count type
* [ ] improving container and switch ?
* [ ] comments #56 : markdown ?

rom1504:
See #59 and #60

rom1504:
and #68

roblabla:
Closed in favor of #60 and #167

PR #56, opened 2015-08-31
[WIP] Add protocol_comments.json

roblabla:


rom1504:
To fix : 
* [ ] before/after (should be fixed in my pageToSection function)
* [ ] what to do about the wikitext ? is it okay to store formatting in a json file ?
* [ ] think about how to do notes (is an {"fieldName":"comment"} object possible ?)
* [ ] might be interesting to put in common stuff from my protocol extractor with yours

roblabla:
- yes
- markdown
- fieldName: comment, I could make that work __only__ for outermost fields (so no comments on arrays)
- Yup

rom1504:
That would mean no comments on array and containers yeah, only comments on leaves. Do we need comments on container and array ? (wiki.vg doesn't have them afaik)
If we do, the only way I can see it working is by having the same structure as protocol.json, but I guess that's a bit annoying.

rom1504:
Possibilities we talked about : 
* putting markdown in the .json
* putting reference to markdown files in the .json
* putting reference to the json file in a markdown file

Stuff that need special formatting handling : 
* tables (which I think could be simply put in json files)
* images (just 2 of them, do some  hack there)
* unformatted text
* links
  *  links to other packet : we can probably represent that in a decent way
  * links to types (example : fixed point number) : maybe that would have something to do with the new types field in protocol.json ?
  * links to stuff like objects/entities : that should probably link to that part of the doc (?d=entities)

I think most (if not all) wiki.vg/Protocol tables have their place in minecraft-data as .json
For example http://wiki.vg/Protocol#Particle : just like entities.json
things like http://wiki.vg/Protocol#Player_Digging are mostly description but they could be complemented with more standard names and are useful like that anyway (better have "started digging" than "0")

rom1504:
Stuff to fix anyway here. But it should be 1) rebased on master 2) protocol_comments_extractor.js should go in https://github.com/PrismarineJS/mcdevs-wiki-extractor

rom1504:
Erf, deleting the branch automatically closed it. This is still WIP and will be tracked in https://github.com/PrismarineJS/minecraft-data/issues/50 for now.

PR #66, opened 2015-09-10
Implement the packet changes of snapshot 15w36d

roblabla:


rom1504:
Better make sure this work in nmp before merging I think (related to https://github.com/PrismarineJS/node-minecraft-protocol/pull/212 )

rom1504:
Well let's merge, it can always be corrected if there are some errors when testing in nmp.

PR #73, opened 2015-10-12
Remove unnecessary line

PangeaCake:


rom1504:
Hmm, it actually does something: it removes all lines after "};". (that's not a very clean way to do it, granted, but it works)

PangeaCake:
Oh, I made a mistake. I thought the previous line of code already covered "};". Sorry for the confusion.

PR #79, opened 2015-11-04
Added pitches.json

demipixel:


phase:
I think "num" sounds a little weird. Maybe something like "id" would be clearer.

rom1504:
It's not id, it's "number of uses", whatever that means http://minecraft.gamepedia.com/Note_Block#Usage

Gjum:
It's how often you click on a noteblock to get that note, so it's kind of an offset or ID. `num` is really ambiguous.

rom1504:
So apparently all this is computable https://www.npmjs.com/search?q=note
pitch = 0.5 * (2^(1/12)) ^ num
octave = floor(num / 12)
num=log(pitch)/log(0.5 *  (2^(1/12)))

http://www.phy.mtu.edu/~suits/NoteFreqCalcs.html

Does that mean we don't need this file or should we keep that file and generate it ?

Gjum:
The only thing of that calculation that is specific to minecraft is the fact that `pitch = 1` means `num = 12` means F# (probably need the absolute pitch somewhere), everything else can be calculated from there.

This means it is possible to have at least that info in minecraft-data and let the wrappers provide conversion functions.

On the other hand, having `pitch`, `num`, and maybe `absolute pitch` in minecraft-data would be OK too, because there are just 25 entries anyway and they represent all the data needed to play noteblock music.

demipixel:
None of this data ever changes so, while it would be convenient to have a generator, it would only ever be used once.

Gjum:
@demipixel that's the exact reason for it to be in a separate package, as it is not minecraft dependent.

rom1504:
@demipixel so do you think we still need this data even though https://github.com/mhsjlw/flying-squid/blob/master/src/lib/serverPlugins/sound.js#L34 works ?

demipixel:
It was just useful to have the keys and stuff, but I can't say it was 100% necessary.

(Although sounds and particles would be nice but that's for another time :3)

rom1504:
So, we didn't need that in flying-squid. And it seems everything there can be computed easily.

Closing, if we happen to need it later, the code will always be there (although it really does seem it's not needed)

PR #84, opened 2015-11-14
Add status/potion effects

1b8:
Added potion/status effects as `effects.json` in this format:
```json
{
  "id": 4,
  "name": "MiningFatigue",
  "displayName": "Mining Fatigue",
  "type": "bad"
}
```
I'm not sure if I got the `name`s completely right, and I might change `"type": "bad"` to something else.

### DIFF
--- a/data/1.8/effects.json
+++ b/data/1.8/effects.json
@@ -1,0 +1,140 @@
+[
+  {
+    "id": 1,
+    "name": "Speed",
+    "displayName": "Speed",
+    "type": "good"
+  },
+  {
+    "id": 2,
+    "name": "Slowness",
+    "displayName": "Slowness",
+    "type": "bad"
+  },
+  {
+    "id": 3,
+    "name": "Haste",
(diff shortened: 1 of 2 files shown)


rom1504:
Nice. We can complete it with actual information about effects later if needed (that kind of stuff http://minecraft.gamepedia.com/Status_effect)
Out of curiosity, do you plan on using it for something ?

1b8:
I'm making an MCEdit/WorldEdit schematic-reading library that no one will use, with way too many features. :P

rom1504:
@1b8 you can now use it with (node-)minecraft-data 0.16.0 (that I just released), see https://github.com/PrismarineJS/node-minecraft-data/blob/master/doc/api.md#effects and https://github.com/PrismarineJS/node-minecraft-data/blob/master/example.js#L9

ah fun project, you might be interested in https://github.com/wtfaremyinitials/minecraft-schematic even though it wasn't much advanced I think.

PR #86, opened 2015-12-31
Fix wrong slot indices in brewing stand window

Gjum:
thanks to @pokechu22 for the images from his clientside mod https://imgur.com/a/jPUs1

### DIFF
--- a/data/1.8/windows.json
+++ b/data/1.8/windows.json
@@ -108,11 +108,11 @@
     "slots": [
       {
         "name": "ingredient",
-        "index": 0
+        "index": 3
       },
       {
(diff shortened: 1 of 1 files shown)


rom1504:
I bet the same is true for 1.9 ?

Gjum:
yup. but the brewing stand changes in 1.9 and I don't know the slot nr of the blaze powder yet.

Gjum:
Also the player inventory gets a new slot. So I think all that 1.9 stuff will be another PR then.

rom1504:
Okay, merging.

PR #88, opened 2016-01-05
Clean up protocol.json formatting a bit

oldmanmike:
Some minor formatting inconsistencies I noticed while using protocol.json as a reference.


rom1504:
Thanks !

PR #89, opened 2016-01-07
Fix furnace window properties

Gjum:
pretty sure it's unchanged in 1.9, but should be verified

if you wait a bit with the merge, I migh add other wrong properties if I find them


rom1504:
okay tell me when it's ready to merge ;)

Gjum:
verified all properties for 1.8, ready for merge I'd say

PR #90, opened 2016-01-10
new entities.json file, extracted using node-minecraft-extractor 

rom1504:
from mcdevs and mcwiki

about #72 

@Gjum can you look at this ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ Projects that provide data:
 | Project name | Language | Source | Data provided |
 | --- | --- | --- | --- |
 | [minecraft-wiki-extractor](https://github.com/PrismarineJS/minecraft-wiki-extractor) | Node.js | [minecraft wiki](http://minecraft.gamepedia.com) | blocks, items, entities, recipes |
-| [mcdevs-wiki-extractor](https://github.com/PrismarineJS/mcdevs-wiki-extractor) | Node.js | [wiki.vg](http://wiki.vg/Protocol) | a partial protocol |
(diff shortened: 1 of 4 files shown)


Gjum:
There are two `Item Frame`s (IDs 2 and 71), but no `Dropped Item`s in the new json files. I remember `Dropped Item` being ID 2, so check your extractor on that.

Also, there should be some kind of sorting so the diff doesn't blow up each time something changes, and that humans can look up entries quickly.

rom1504:
It is sorted, just not the same order as before.
It currently puts the mob first and then the object. Appart from that it's sorted by id.
(it's used to be completely sorted by id which wasn't making that much sense)

I fixed the dropped item thing.

Gjum:
LGTM

metadata can be a different PR if you want.

Gjum:
On further inspection there still seem to be some issues:
`curl https://raw.githubusercontent.com/PrismarineJS/minecraft-data/new_entities/data/1.8/entities.json | grep -oP '(?<=displayName": ")[^"]*' | sort | uniq -c`
gives `3 Minecart` and `2 Falling block`

rom1504:
Yes but I'm not sure what to do about this. The minecart thing is due to wiki.vg having old entities (see http://wiki.vg/Entities#Objects) which mcwiki doesn't have, and I'm taking the names from mcwiki.

Same thing about falling block, mcwiki says it's the same entity ... (http://minecraft.gamepedia.com/Data_values/Entity_IDs : falling block has sand and dragon)

PR #91, opened 2016-01-10
wiki.vg displayName

rom1504:


rom1504:
going with mcwiki displayNames

PR #92, opened 2016-01-24
Add protocol version data

deathcap:
Adds the data from http://wiki.vg/Protocol_version_numbers

* data/common/protocolVersions.js: object mapping release version to protocol version
* data/common/majorVersions.js: object mapping release version to "major" version, so it can be used to index into the minecraft-data datasets (which currently only support one major version each)


deathcap:
Restructured to use an array of objects instead, to match the style of minecraft-data. This also includes adding a new `useNetty` boolean field instead of overloading `protocolVersion`:

```json
  {
    "releaseVersion": "16w03a",
    "protocolVersion": 96,
    "usesNetty": true,
    "majorVersion": "1.9"
  },
…
  {
    "releaseVersion": "1.0.0",
    "protocolVersion": 22,
    "usesNetty": false,
    "majorVersion": "1.0"
  }
```

rom1504:
Nice, thanks for the pr

Issue #96, opened 2016-02-07
Light emitting values of blocks


rom1504:
@hansihe was saying in #mcdevs that http://minecraft.gamepedia.com/Light#Light-emitting_blocks are interesting values
Possibilities:
* putting it in blocks.json, but it would mean expanding the extractor https://github.com/PrismarineJS/minecraft-wiki-extractor/blob/master/lib/block_extractor.js again
* putting it in a blocks_lights.json because it's kind of separate from most uses of blocks

(it would anyway be exposed as a single block thing in wrappers)

rom1504:
A benefit of blocks_lights.json is it could be written manually to begin with ;)

rom1504:
fixed by #102

nickelpro:
Right now light emitting blocks such as lava are listed as "transparent" when that's not really true. They're opaque light emitters in their own right, not transparent to the light level above them.

rom1504:
Hmm the wiki do list lava as transparent http://minecraft.gamepedia.com/Lava

@hansihe how would you handle lava in your light handling thing ?

(also see WIP in https://github.com/PrismarineJS/minecraft-wiki-extractor/issues/12 )

hansihe:
I do think what @nickelpro is saying is technically true, but I don't think it realistically matters since lava emits the highest possible light level, 15.

hansihe:
Lava would be handled just like a normal light emitting block. Since light emitters don't affect blocks that are lit up to a higher light level by another light emitter, and lava emits the highest possible light level, 15, I don't think there is any possible way changing the filterlight of lava would make any difference.

Gjum:
This could be verified by creating a closed room with only lava (/glowstone/redstone lamps) as walls and looking at the skylight in the F3 screen.

hansihe:
@Gjum has a point, what I said earlier was false. If I'm not mistaken, the filterlight value of lava would in fact affect skylight.

PR #99, opened 2016-02-13
Add block_lights.json (1.8 only for now)

hansihe:
Adds block_lights.json which contains data on what blocks emit and filter light. This along with transparency data for each block (not currently in blocks.json, that is another issue) should be enough to calculate lighting in a Minecraft world.

Solves #96.


rom1504:
we prefer mcdata schemas to be array so that wrapper/reader can index it how they want easily (https://github.com/PrismarineJS/minecraft-data/issues/100)
So I think it would be better to have something like that:
```json
[
{
"blockId":138,
"emitter":15
},
{
"blockId":60,
"filter":15
}
]
```

(you can probably transform your file to look like that automatically)

hansihe:
Right, I wasn't aware any schema was preferred, I should probably have researched a bit more before writing it.

Will make it more like the format you proposed.

rom1504:
would be better to name it `id` than `blockId` actually, since we name it id in other files.

rom1504:
extracted automatically and put in blocks.json in #102

PR #102, opened 2016-02-13
add emitLight and filterLight

rom1504:
not sure if the transparent property is still useful ?


rom1504:
(transparent == false is filterLight==15 and transparent == true is filterLight == 0)

Issue #27, opened 2015-08-07
Packets and field names


rom1504:
Currently packets names are taken from https://github.com/aadnk/ProtocolLib/blob/master/ProtocolLib/src/main/java/com/comphenix/protocol/PacketType.java#L123
I don't know where the fields names are coming from (@roblabla ?)

These names are different from what is currently found in http://wiki.vg/Protocol

Can reasonable names be chosen to make protocol.json more standard ?

roblabla:
Field names are mostly taken from wiki.vg, ProtocolLib doesn't have any for them.

I think using packet names from ProtocolLib is still a better idea, than coming up with our own, y'know, wheel reinvention and all, it ain't great.

It might be worth going over all the fields and remove the discrepencies (like having entityId in some places and id in some others, for instance).

It's worth noting that if this is a problem stemming from having an automatic generator, it'd be easy enough to parse ProtocolLib's sources to find the name associated with a packetid...

rom1504:
Okay so field names were originally taken from wiki.vg, but then they were probably changed because a lot of them are different from wiki.vg (see https://github.com/PrismarineJS/minecraft-data/blob/master/bin/wiki_extractor/protocol_extractor.js#L513 )

It's not so much about coming up with our own, but the fact is have there is now 2 standard : wiki.vg and ProtocolLib (and probably so many others in the list in that gist). If we are going to use protocol.json to produce something like wiki.vg/Protocol, we need to either take wiki.vg names, either push ProtocolLib names to wiki.vg. An other way might be to keep different names for every packet, but I'm not sure that's a great idea.

Yeah indeed there are some discrepancies in field names that should be fixed (like gamemode vs gameMode)

It's not a big problem from the generator side : I actually already fixed that by having a id to ProtocolLib names converter. I mean it's not a technical problem. Also the main point of my wiki generator is to be able to see the difference between the wiki and our protocol.json : to see if some types are wrong for example. The second thing I want to use it for is getting field notes into protocol.json. I'm almost there.

In a second time, that generator should be rendered useless by going the other way (protocol.json to something like wiki.vg/Protocol)

nickelpro:
+1 for wiki.vg because that's what SpockBot uses and I do not like change, plus the names are more descriptive (Collect Item vs COLLECT, Set Experience vs EXPERIENCE, Entity Properties vs UPDATE_ATTRIBUTES)

[...]

PR #105, opened 2016-02-15
16w04a update

oldmanmike:
As far as I my eyes can tell, protocol.json is now synced up with the [pre-release protocol page](http://wiki.vg/Pre-release_protocol). This includes the names of the packets so that developers can grep for packets in the json based on the names used on the wiki. Changing packet names might not be desirable for projects directly using protocol.json and are expecting to packets to maintain a stable name, even if the name is out of sync from the wiki. I can switch it back if it turns out to be more trouble than it's worth.

This pull request does not include the reordering of the packets in protocol.json based on packet IDs. This would be quite a bit of work, but it will help readers orient themselves and could result in spotting duplicate packet ID uses.

### DIFF
--- a/data/1.9/protocol.json
+++ b/data/1.9/protocol.json
@@ -295,7 +295,7 @@
             }
           ]
         },
-        "join_game": {
+        "login": {
           "id": "0x23",
           "fields": [
             {
@@ -328,7 +328,7 @@
             }
           ]
         },
-        "chat_message": {
+        "chat": {
           "id": "0x0f",
           "fields": [
             {
@@ -341,7 +341,7 @@
             }
           ]
         },
-        "time_update": {
+        "update_time": {
           "id": "0x44",
           "fields": [
             {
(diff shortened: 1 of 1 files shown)


rom1504:
Nice!
Indeed I don't think we're ready to use wiki.vg packet names, see that issue https://github.com/PrismarineJS/minecraft-data/issues/27. It would be better to keep that PR as an update to 1.9 and not update the name.

rom1504:
There was some reason for not having packets ordered by packets ids (eg that similar packets (entity packet for example) are kept close), but I think it's probably more trouble than it's worth, and it might be better to reorder them indeed. (but that would be in an other PR)

oldmanmike:
Alright, I'll revert the names.

rom1504:
Alright I tested this with node-minecraft-protocol proxy and there was only one error: held_item_slot is 0x17 not 0x14. That kind of mistake would be easy to spot if packet were in the id order, I think we should move back to that.

Issue #106, opened 2016-02-15
Reorder 1.9 protocol by packet id


rom1504:
It makes updating the protocol easier and it doesn't make that much more sense to try to order the packets in a meaningful order.
see #105

oldmanmike:
I can knock this off while I've still got it on my mind.

rom1504:
fixed by #108

PR #107, opened 2016-02-16
Reordered 1.9 protocol packets

oldmanmike:
Reorders all the 1.9 packets by packet ID as to provide the desired property outline in #106.


oldmanmike:
Yeah, this is pretty messy of me.

PR #109, opened 2016-02-16
Restore support for 15w40b

deathcap:
https://github.com/PrismarineJS/minecraft-data/pull/105 is good in that it updates the 1.9 data to 16w04a, but bad in that it breaks backwards-compatibility (e.g., I already have a 15w40b server instance to test with, but this is an incompatible protocol change which was working before the minecraft-data update). This PR restores the 15w40b protocol in its own separate directory (no other data - can be shared with 1.9).


rom1504:
How would "sharing with 1.9" works ? Use the most recent existing version of a file for a common major version in node minecraft data ?
Does it really makes sense to do that ? (blocks and the rest of the data change a bit between each snapshot, even if I don't see how to get the data by snapshot from the wiki (data by version is doable because there's a big gap between the dates, so using the correct revision does it))

deathcap:
Yes that's what I'm doing in https://github.com/PrismarineJS/node-minecraft-data/pull/17/files, protocol and version are 15w40b but all other data is '1.9'. I think it mostly makes sense, there were at least no non-protocol changes in the 16w04a update at least in https://github.com/PrismarineJS/minecraft-data/pull/105/files?diff=unified - if/when there are significant block/item changes, then I think those could be separated out as well, as needed.

rom1504:
Okay let's do that

PR #116, opened 2016-02-18
1.7 data

rom1504:


rom1504:
ah damn mineflayer never had 1.7 data https://github.com/PrismarineJS/mineflayer/commits/master?page=10

rom1504:
for reference, first node.js commit of mineflayer was at 1.4.6 https://github.com/PrismarineJS/mineflayer/tree/217d1fd3084134a6b4eafd1aae1a0632a6a3839b

rom1504:
Only thing left is recipes, which I'd like to extract.
Problem is my script is based on dpl queries and they don't properly handle a revision time, so I have to rewrite it to read directly the pages, see https://github.com/PrismarineJS/minecraft-wiki-extractor/issues/2#issuecomment-151684303

I'll either do that, or I guess we won't have 1.7 recipes for now.

rom1504:
Well let's merge this, I'll add recipes later.

PR #120, opened 2016-02-19
1.7 custom_payload 0x3f varshort for Forge compatibility

deathcap:
Requires https://github.com/roblabla/ProtoDef/pull/52

http://wiki.vg/Minecraft_Forge_Handshake#Differences_from_Forge_1.7.10 "However, forge makes some more changes to the server to client packet [0x3f](but not the client to server packet [0x17]): Rather than using a short for the length, a varshort is used"

Note this is backwards-compatibile with vanilla 1.7, since short lengths would never be negative. varshorts extend short by adding an extra byte (to extend 16-bit range to 23-bit) following the negative short with MSB set. Tested on forge-1.7.10-10.13.4.1614. Would like this for https://github.com/PrismarineJS/node-minecraft-protocol-forge/pull/7 (couldn't make the change there since it is a lower-level protocol change, but at least this is backwards-compatible with vanilla, so it should be safe - only extends the protocol)

### DIFF
--- a/data/1.7/protocol.json
+++ b/data/1.7/protocol.json
@@ -1,6 +1,7 @@
 {
   "types": {
     "varint": "native",
+    "varshort": "native",
     "string": "native",
     "u16": "native",
     "u8": "native",
@@ -1916,7 +1917,7 @@
               "type": [
                 "buffer",
                 {
-                  "countType": "i16"
+                  "countType": "varshort"
                 }
               ]
             }


rom1504:
This can now be moved to nmp-forge using https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_custom_packets/client_custom_packets.js (expose a customPackets in nmp-forge probably to be used along createClient probably ? maybe there should be a better way to install this but I'm not sure what)

deathcap:
Hmm I can't see how I can use the `customPackets` option here, nmp-forge needs to change the custom packet definitions after it has already been created because the `forgeHandshake` and `autoVersionForge` plugins augment node-minecraft-protocol createClient after setup, but by this time the serializers/deserializes have already been created.

rom1504:
@deathcap what about providing a require('minecraft-protocol-forge').customPackets ?

deathcap:
```json
        "custom_payload": {
          "id": "0x3f",
          "fields": [
            {
              "name": "channel",
              "type": "string"
            },
            {
              "name": "data",
              "type": [
                "buffer",
                {
                  "countType": "varshort"
                }
              ]
            }
          ]
        },
```

in `customPackets` should work, but the problem is where to add it. 

src/client/forgeHandshake (the node-minecraft-protocol "plugin"), exports this function to augment `client` for Forge support:

```
module.exports = function(client, options) {
  client.tagHost = '\0FML\0';
  client.on('custom_payload', function(packet) {
    if (packet.channel === 'FML|HS') {
      fmlHandshakeStep(client, packet.data, options);
    }
  });
};
```

would like to also augment it to change the custom_payload 0x3f protocol definition. By this time, the constructor for createClient() has already been called.

Trying to access `client.serializer.proto` but not clear how/if it can be modified to change this definition. There is a `packet` type but it only contains the read, write, and sizeOf functions. Is it possible to get to the packet types directly through `client.serializer.proto`?

rom1504:
@deathcap I really don't understand what is the problem with doing this:
```js
var mc = require('minecraft-protocol');
var forgeHandshake = require('minecraft-protocol-forge').forgeHandshake;
var client = mc.createClient({
    host: host,
    port: port,
    username: username,
    password: password,
    customPackets: require('minecraft-protocol-forge').customPackets
});

forgeHandshake(client, {forgeMods: [
  { modid: 'mcp', version: '9.18' },
  { modid: 'FML', version: '8.0.99.99' },
  { modid: 'Forge', version: '11.15.0.1715' },
  { modid: 'IronChest', version: '6.0.121.768' }
]);
```

Also see https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_custom_packets/client_custom_packets.js#L8
The format of custom packets has changed (same as in mcdata protocol.json)

rom1504:
(if you really want to modify the serializer after it has been created you could expose and call createProtocol (https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/transforms/serializer.js#L19) and then change the protocol in serializer)

rom1504:
Oh and customPackets are merged, so for example https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_custom_packets/client_custom_packets.js#L8 adds a packet, it doesn't remove them all.

rom1504:
tracked elsewhere

PR #122, opened 2016-02-22
add packet type in protocol.json

rom1504:
solve https://github.com/PrismarineJS/node-minecraft-protocol/issues/301

Makes it possible to do https://github.com/mhsjlw/minecraft-classic-protocol/blob/master/src/transforms/serializer.js#L11

instead of https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/transforms/serializer.js#L17

I think it's better that way.

Only thing I'm not sure about is the global types properties being called "types". Since it's all type now.


rom1504:
I'd like to have a "namespace" concept instead of the state, direction ones to make the schema more general. (I would like to handle stateless protocols with the same schema definition, for example https://github.com/mhsjlw/minecraft-classic-protocol/blob/master/data/protocol.json )

If anybody has any idea how a namespace schema could be done, please share.

rom1504:
To do:
* [ ] update the doc (mostly that part https://github.com/PrismarineJS/minecraft-data/blob/gh-pages/loadProtocol.js#L35 , make it recursive)
* [ ] adapt nmp
* [ ] make a small module to read the .json and add provide protocols by namespace (or add it directly to protodef)

rom1504:
About schema improvement : think about where datatypes should go. Having entityMetadataLoop and entityMetadataItem in the schema that validates the classic protocol is kind of wrong.
Instead of including every datatype imaginable inside that schema, it would be possible to just say that a datatype is a string or an array with first element being the type name, and second element being any object or array.
Then there could be a second validation to make sure these types make sense relatively to what's added to the protodef instance. That should still happen during the testing phase, doing this at runtime is too slow.
Not 100% sure how this would be done, but it's related to taking the datatype schema out of that big protocol definition, which I want to do for https://github.com/roblabla/ProtoDef/issues/46#issuecomment-186217702 (also related : json schema can reference subschema in separate files)

That should probably be done in another PR though. This PR is already doing a lot of changes as-is.

rom1504:
Ah, to justify my namespace approach: the idea is the protocol.json is now an object with a property "types" which contain types, and other properties that are namespaces that contain an object with the same format as protocol.json. That leaves protocol users define as many level as needed.

PR #132, opened 2016-04-05
entity_metadata.json for 1.8

hansihe:
This is still largely unverified/untested. It would be nice if someone could look it over.

The format is mostly made up. Please comment if you have any improvements on it.

### DIFF
--- a/doc/history.md
+++ b/doc/history.md
@@ -1,3 +1,6 @@
+## 2.15.0
+* supports 17w18b
+
 ## 2.14.0
 * supports 17w15a
 


Gjum:
There was some talk about the format in #72 already, the major difference being that the fields are objects instead of arrays, like this:
```json
"fields": [
 { 
   "index":17,
   "type":"int",
   "meaning":"Shaking Power"
  }
]
```
I personally prefer objects over arrays here because of extensibility and readability.

rom1504:
Yes I agree with @Gjum 
Also the bitmasks values are missing (for example http://wiki.vg/Entities#Guardian )
It could be done with a format like `[{"bitMask":"0x02","displayName":"is retracting spikes","name":"isRetractingSpikes"}]` but we don't have to do that for the first pass of that file.

rom1504:
oh why delete this ?
well I kind of understand, it was in master branch, but could you get back the commits, put it in another branch and re-open a pr ?

PR #137, opened 2016-05-23
Fix minecraftVersion in 16w20a

oldmanmike:
My generator stumbled upon a typo.


Gjum:
:+1: this should be checked in auditing

Issue #65, opened 2015-09-10
MCPE support


roblabla:
It'd be interesting to have the MCPE data available here as well. I believe the two games are close enough to actually be able to support both schema types. The protocol will probably be the one file to change the most, but I believe its schema won't have to change all that much either

rom1504:
Yeah I agree , I guess we would follow the same organization than versions (so branches for now and possibly folders if we do choose to use folders in #64 )

rom1504:
It would be nice to find some good sources to build a protocol.json for mcpe.

roblabla:
@rom1504 there's also PocketBurger https://github.com/PocketMine/PocketBurger

rom1504:
Also see where @Creeplays is taking his data from (https://github.com/mhsjlw/mcpe-server/blob/master/lib/net/protocol.js#L75)

rom1504:
http://minecraft.gamepedia.com/Pocket_Edition : ah the last version is only available in windows 10

mhsjlw:
@rom1504 Don't use the original source for the mcpe-server, we've updated the packet names so the old Protocol.js file is no longer usable

rom1504:
What about https://github.com/PocketMine/PocketMine-MP/tree/master/src/pocketmine/network/protocol ? seems more or less up to date

mhsjlw:
@rom1504 Looked at it a while ago, it is useless

mhsjlw:
This... now this... is what is useful https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/network/protocol/Info.php

rom1504:
No, that second file give the packets id<-->name, but that's not very useful, the first link give you all the fields for each packet.

mhsjlw:
@rom1504 Can we turn this into documentation?

rom1504:
We can probably turn it into a protocol.json, which can then be converted automatically to a doc.
It's the 0.12 branch we'd be more interested about https://github.com/PocketMine/PocketMine-MP/tree/mcpe-0.12/src/pocketmine/network/protocol

rom1504:
https://github.com/rom1504/mcpe-protocol-extractor

rom1504:
I guess we'll be following the same structure as currently : a new branch for each mcpe version.

So that means a mcpe-0.12 branch.

It's related to #64 . Continuing this comment there.

phase:
What if this was put in another repository, like *mcpe-data*?

[...]

PR #138, opened 2016-05-23
Minecraft pocket edition data : blocks and items

rom1504:
I put the data in data/pe.
I think the rest of the data should move to data/pc.
That will mean all the wrappers will need to update their code when updating the submodule though.


rom1504:
related : #65

rom1504:
Okay I moved pc data to pc/
Will merge in a bit, and release this as 2.0.0 (and then adapt the wrappers)

PR #142, opened 2016-06-07
split the protocol schema

rom1504:
one file by type, first step toward validating the datatypes in protodef and extensibility of the protocol schema

the schema files will probably go in https://github.com/ProtoDef-io/ProtoDef

Using this it can be possible to do proto.addType(name,definition,schema) and to automatically validate the type. The schema parameter can be made optional simply by adding an all validating schema for this datatype when not passed. Most types schema will simply be "dataType" (see datatype.json) , i.e. one of the already added types.

see https://github.com/ProtoDef-io/node-protodef/issues/46


roblabla:
Hmm, what's the advantage of the schema over inferring types from the existing "type" ? My current idea was to add a fourth function to types (`validate` ) that would make sure the required typeArgs are there, valid 'n all.

If we go down this route, I'd much rather the schema is added to the read/write/sizeOf array.

rom1504:
What do you mean by `inferring types from the existing "type" ` ?
I just added a sentence to my post : Most types schema will simply be "dataType" (see datatype.json) , i.e. one of the already added types.
Maybe that answer that question ?

I like better schemas vs a validate function because they are descriptive, you can use them to know how a type definition should look like. And I also think they can be easier to write than a function.

> I'd much rather the schema is added to the read/write/sizeOf array.

What's the point of doing that ?

I'd just thought about having an ajv instance and validating types using it. The schemas are independant from js that's why they would make sense in the protodef spec.

roblabla:
The reasoning is, the validation is a part of the definition. I don't see the use in keeping them separate. The idea is that a datatype can easily be published to npm as an X-element array. Then it can be named however the user wants.

A validation function is a lot more flexible. It'd allow meta-programming facilities that may be useful, such as traversing the JSON, etc. I feel like schemas are very restrictive in what they allow us to do.

Isn't it possible to implement schema validation as a kind of "validate function" ? I imagine the validate function would take `(typeArgs, cb(fatalerr, errList))` ? (fatalerr being an error outside the validation).

Here's my thought : let's take the countType field. That field requires a reference to a "number". What'd be really nice is to think up a scheme to let a field tell its parent it returns a "number" so the countType is valid. This level of validation would allow to catch lots of errors really quickly. I don't think schema validation can catch those errors.

rom1504:
> The reasoning is, the validation is a part of the definition. I don't see the use in keeping them separate.

But the definition is in the spec, the read/write/sizeOf function are only one implementation.

> The idea is that a datatype can easily be published to npm as an X-element array. Then it can be named however the user wants.

Ah yeah for non standard types, the schema/validation will probably go next to the implementation, especially if other implementations don't use it.

Yeah I guess we could allow both validation functions and schemas.

> Here's my thought : let's take the countType field. That field requires a reference to a "number". What'd be really nice is to think up a scheme to let a field tell its parent it returns a "number" so the countType is valid. This level of validation would allow to catch lots of errors really quickly. I don't think schema validation can catch those errors.

I think this belongs to the compilation phase. Maybe elixir-protodef already does that btw. ( @hansihe ?) It requires to know all the context of the type.

But anyway I guess we could have both. Currently we have schemas that catch some errors already, but we could indeed have more precise validation functions that would catch more things.

hansihe:
elixir-protodef does not currently do that. I tried doing it, but I could not think of a flexible way for a type to indicate what sort of data it takes/returnes, and put it off for later. The few cases where that is done is currently handled at runtime.

I would obviously prefer if things where kept relatively language agnostic, but I am not sure what the best way of doing that would be. As is elixir protodef should only compile protodef files that actually work (with the exception of type mismatches, as mentioned above), so I suppose it already does quite a bit of validation and sanity checking.

rom1504:
* https://github.com/ProtoDef-io/ProtoDef/issues/9
and
* https://github.com/ProtoDef-io/node-protodef/issues/77

move this issue forward a bit

Required next : 
* use the data in ProtoDef to validate the schema here. To do that have a validation function somewhere (directly in node-protodef or in a module used by node-protodef and minecraft-data)
* take into account the more general validation function remark to improve the implementation in node-protodef

PR #146, opened 2016-08-20
Add 0.15 Minecraft: PE protocol data

mhsjlw:
This is a pull request to add Minecraft: PE 0.15.6's protocol data to minecraft-data. It would be great if @rom1504 ran his generator on the 0.15 version of the wiki page to update all of the blocks, items, etc. so for now I just added the protocol. It is quite stable as it was used to get both a player and a server to join and host a game (respectively).


rom1504:
hmm does this really work ? I don't remember where we were at about this.

rom1504:
Yeah https://github.com/mhsjlw/pocket-minecraft-protocol/issues/14 says it works. Sorry for taking so much time to merge this.
Will publish a new node-minecraft-data version with this later today.

mhsjlw:
Thanks :+1:

Issue #157, opened 2017-01-24
skeleton horses are missing


plexigras:
https://prismarinejs.github.io/minecraft-data/?d=entities
http://minecraft.gamepedia.com/Data_values/Entity_IDs

rom1504:
is that a 1.11 thing ?

rom1504:
Well anyway, I didn't extract the 1.11 data, should do that.
Related to #155

plexigras:
http://wiki.vg/Protocol_History#16w32a
> Zombies, skeletons, and horses have been split into their component entities rather than having a Type field (although the Type field still exists on Zombies, it is unused).

plexigras:
i found this because in the current npm version(2.0.0 but not in git master) mineflayer crashes

so might wanna push that to npm

rom1504:
No the 1.11 data is not even in mcdata yet.
However, you could try the last commit of mineflayer, it should prevent crash due to entities ;)

rom1504:
fixed https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.11/entities.json#L272

Issue #155, opened 2016-11-27
1.11 changes to entities


Pokechu22:
* `Mob` and `Monster` don't exist any more
* IDs are lowercase and namespaced, and some changed (`minecraft:tnt` vs `PrimedTnt`)
* There are no objects with ID 11 and 12 any more (and that hasn't been the case for a while, I think)
    * All minecarts use ID 10, so that's going to need to be represented some way (it's specified in the object data)
* Some sizes are wrong
* Horses!  You're missing the different types of horses...  and `EntityHorse` is no more (even with legacy names); it's `Horse` now.

rom1504:
I didn't extract the 1.11 entities file at all yet actually.

It is extracted using both mcwiki and wiki.vg, I'll do it. (hopefully http://wiki.vg/Entities is up to date)

(using https://github.com/PrismarineJS/node-minecraft-extractor )

rom1504:
I believe this is all fixed. see https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.11/entities.json

PR #158, opened 2017-01-26
add 1.11 entities

rom1504:
using https://github.com/PrismarineJS/node-minecraft-extractor/commit/9f2be44744466f7a6203fa9d6a80dce9077aa7f3

for #155 and #157


rom1504:
will prob extract the rest before merging.

But if anyone want to check the data, would be useful

PR #165, opened 2017-03-12
Fix incorrect ID for clientbound custom_payload

Pokechu22:
This was my mistake: when editing the article to add string max lengths, I changed the packet ID to 0x20 instead of putting a length of 20 to the channel ID.


rom1504:
thanks !

PR #171, opened 2017-03-31
Air should not be diggable (pc/1.11)

iRath96:
[data/pc/1.11/blocks.json](https://github.com/PrismarineJS/minecraft-data/blob/bf4302f39b74c2fcc23d840bc4a2eb57a0262797/data/pc/1.11/blocks.json#L8) incorrectly specifies that air blocks are diggable.

### DIFF
--- a/data/pc/1.11/blocks.json
+++ b/data/pc/1.11/blocks.json
@@ -5,7 +5,7 @@
     "name": "air",
     "hardness": 0,
     "stackSize": 0,
-    "diggable": true,
+    "diggable": false,
     "boundingBox": "empty",
     "drops": [
       {


roblabla:
Those are automatically generated from [the extraction scripts](https://github.com/PrismarineJS/minecraft-data#extraction).

iRath96:
@roblabla Thanks for your response. I've run the extraction script myself now by doing this:

1. `git clone https://github.com/PrismarineJS/minecraft-wiki-extractor`
2. `cd minecraft-wiki-extractor`
3. `npm install`
4. `node bin/block_to_json.js blocks.json existing_items.json existing_blocks.json`

with `existing_items.json` and `existing_blocks.json` being `[]`.

This yields correct behavior for me (`diggable: false` for the air block), so it seems like the extraction scripts are working correctly 👍 . (that is unless I invoked them incorrectly)

roblabla:
@rom1504 we might want to run the scripts to update this repo then ?

PR #176, opened 2017-05-10
add PDS file for 1.11

hansihe:
Adds an initial converted PDS file for 1.11.

There are still a few things in the file that are slightly hackish due to missing features in protodefc, but I think this is an acceptable initial state.

### DIFF
--- a/data/pc/1.11/protocol.pds
+++ b/data/pc/1.11/protocol.pds
@@ -1,0 +1,1126 @@
+@doc "8-bit signed integer"
+@type integer("i8") def_native("i8");
+
+@doc "8-bit unsigned integer"
+@type integer("u8") def_native("u8");
+
+@doc "16-bit signed integer"
+@type integer("i16") def_native("i16");
+
+@doc "16-bit unsigned integer"
+@type integer("u16") def_native("u16");
+
+@doc "32-bit signed integer"
+@type integer("i32") def_native("i32");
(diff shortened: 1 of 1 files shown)


rom1504:
Is the PDS format documented somewhere ?

hansihe:
As of yet, no.

rom1504:
so the .pds is in mcdata, but that doesn't make it magically used by things.
I guess node-minecraft-data could provide the file content as a big string.
But then we need https://github.com/hansihe/protodefc/issues/10

rom1504:
@hansihe btw, is there now a script to convert a .json to a .pds ? (possibly taking the doc from somewhere else)
We probably don't want to transform everything manually.

Also I think it would be better if the .json and the .pds contained the same information.

How does the protodefc json thing look like now ?

hansihe:
[This](https://gist.github.com/hansihe/ca5f4c2d0e69691c0c05da626cf8ada2) is an example of a json spec file. The format is still very loosely defined, and it doesn't contain all the same information as a .pds file does yet, but that's planned. It is currently used by the documentation generator only.

The main thing making it hard to automatically translate a protocol.json from the current format is the semantic difference between switch/mapper and union. If I worked on it, I could definitely get it to work fairly well, but I am unsure how much time I should spend on that versus other things. What do you think?

PR #181, opened 2017-06-20
extract data for 1.12 and copy the rest from 1.11

rom1504:
Related : https://github.com/PrismarineJS/mineflayer/issues/533

I think there are a few things to fix in the extractor, I'll merge after that.

Feel free to look if everything is there


rom1504:
@VTCAKAVSMoACE is everything there ? (in particular in blocks.json)

rom1504:
I think it's good enough, let's merge

Issue #78, opened 2015-10-29
images ?


rom1504:
see https://github.com/wurstmineberg/assets.wurstmineberg.de/tree/master/img/grid

It might be interesting for http://prismarinejs.github.io/minecraft-data purposes

Gjum:
What about just extracting them from the `.jar` when generating the docs?
I don't feel the need to have these in minecraft-data, as they are more of a (default, preinstalled) resource pack.

Also, legal issues should be considered before we redistribute them.

rom1504:
Yeah it's true that these images take 3MB.

Maybe they should be stored elsewhere (the mcdata docs are hosted by github though, so that would be somewhere else on github (/an other repo))

No clue about legal issues but http://minecraft.gamepedia.com/Data_values/Block_IDs is using these images just fine.

Gjum:
[mojang's brand guidelines](https://account.mojang.com/documents/brand_guidelines) about "assets":

> [who is using them] MUST [...] NOT include anything else around it that makes people think that you or what you are doing is official; approved or endorsed by us; or associated or connected with us;
> [...]
> sufficiently differentiate the use of any Name you are using from any other branding.

(last part regarding the PrismarineJS name)

So using them in the docs with a [disclaimer](http://minecraft.gamepedia.com/File:TNT.png?version=03a4cf2c3e0955aa17d72e1f070037d2) should be fine.

rom1504:
On that theme https://github.com/deathcap/ProgrammerArt is interesting, if not vanilla mc.

roblabla:
Well, it's already been mentioned, but I think this repo should keep using as much of the mojangian data as it possibly can. Given the guidelines (and the track record of Mojang), I think it'd be safe to include their assets, properly categorized, if we put a disclaimer somewhere in the README.

deathcap:
:-1: for including Minecraft image assets in minecraft-data, not only would bloat the repo unnecessarily but a better solution imho is for graphical clients to support loading arbitrary resource/texture packs (e.g. using [this module](https://github.com/deathcap/artpacks)), which can include minecraft.jar if the user chooses, or a freely-redistributable resource pack like my own [ProgrammerArt](https://github.com/deathcap/ProgrammerArt) (CC), or [Pixeludi Pack](https://github.com/SirCmpwn/TrueCraft#assets) as used in TrueCraft

[...]

PR #183, opened 2017-06-23
add language files

rom1504:


rom1504:
question : 
* should it be called en_us.json ?
* what about all the other languages ?

rom1504:
or should it go to https://github.com/rom1504/minecraft-assets ?
I have no idea

roblabla:
What's the purpose of minecraft-assets compared to minecraft-data ? Couldn't they be fused ?

rom1504:
see https://github.com/PrismarineJS/minecraft-data/issues/78

Currently minecraft-assets contain json files to describe models and textures and the corresponding images (straight from the client jar).
That means it's a bit heavy (a few MB) and it contains copyrightly dubious content.

It's probably best to keep the images in a separate repo.

But I guess the content that is transformed a bit and put as json can go to minecraft-data.

Including that language.json file.

PR #189, opened 2017-08-03
Updates for 1.12.1

DeliciousJaffa:
Incomplete work for 1.12.1

Not sure what to do with the following new packets:
Play clientbound 0x2B | Unknown (clientbound)
Play serverbound 0x12 | Unknown (serverbound)

Or removal of:
Play serverbound 0x01 | Prepare Crafting Grid


rom1504:
Thanks !
I think we'll need to wait a bit to know what are 0x2B and 0x12

PR #193, opened 2017-09-19
Update protocol.json

lefela4:
There is little fix on some server

### DIFF
--- a/data/pc/1.12.2/protocol.json
+++ b/data/pc/1.12.2/protocol.json
@@ -589,6 +589,14 @@
             }
           ]
         ],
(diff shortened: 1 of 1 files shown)


rom1504:
Node Minecraft protocol custom packets functionality should be used for that. Or fixed if it doesn't work.

roblabla:
This isn't an official minecraft packet. I'm closing this. If you need to run this with NMP, look at https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/examples/client_custom_packets/client_custom_packets.js

For other users of minecraft-data, you'll need to ask their respective authors

PR #195, opened 2017-12-28
Add 1.2 data

mhsjlw:
This is a little more lacking then I'd like it to be.

Generated by https://github.com/mhsjlw/minecraft-bedrock-data a.k.a. sel-utils

@rom1504 if you could run the wiki-scraper and maybe fill up some of the empty spots, that would be appreciated

Protocol data is currently in the works

### DIFF
--- a/data/pe/1.2/biomes.json
+++ b/data/pe/1.2/biomes.json
@@ -1,0 +1,443 @@
+[
+	{
+		"id": 0,
+		"color": null,
+		"name": "Ocean",
+		"rainfall": null,
+		"temperature": 0.5
+	},
+	{
+		"id": 1,
+		"color": null,
+		"name": "Plains",
+		"rainfall": null,
+		"temperature": 0.8
+	},
+	{
+		"id": 2,
+		"color": null,
+		"name": "Desert",
+		"rainfall": null,
+		"temperature": 2
+	},
+	{
+		"id": 3,
+		"color": null,
+		"name": "Extreme Hills",
+		"rainfall": null,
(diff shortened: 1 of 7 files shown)


rom1504:
can you complete : 
* https://github.com/PrismarineJS/minecraft-data/blob/master/data/dataPaths.json
* https://github.com/PrismarineJS/minecraft-data/tree/master/data/pe/common
?

rom1504:
Also it would be nice to run a JSON.stringify(JSON.parse(content),null,2) on the files, it seems the indentation is wrong.

rom1504:
Ah and (that's optional) could you put https://github.com/mhsjlw/minecraft-bedrock-data in https://github.com/PrismarineJS/minecraft-data#extraction

Also how do you run that extractor ? (any readme ?)

mhsjlw:
I'll add a readme, sure. Effectively, grab the xml files from upstream, put them into templates/json then run `make` and run the executable `./src/all` (you must run from src). It will write the json to `./output`

mhsjlw:
Script for others:

```javascript
const fs = require('fs');

files = [ 'biomes', 'blocks', 'effects', 'enchantments', 'entities', 'items', 'version' ];

files.forEach(file => {
  console.log(`Formatting ${file}`);
  const contents = fs.readFileSync(`${file}.json`).toString();
  const formatted = JSON.stringify(JSON.parse(contents), null, 2);
  fs.writeFileSync(`${file}.json`, formatted);
});
```

rom1504:
Why did you close ?

PR #201, opened 2018-08-05
1.13 Support

### LINKED ISSUE #200: 1.13 support
create a 1.13 dir and : 
* [x] update protocol : 
  * last version in mcdata is 17w50a https://github.com/PrismarineJS/minecraft-data/commits/master
  * there's a bunch more versions since : http://wiki.vg/Protocol_version_numbers
  * see history between http://wiki.vg/index.php?title=Pre-release_protocol&oldid=13556 and http://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150 :  http://wiki.vg/index.php?title=Pre-release_protocol&offset=&limit=150&action=history
   * update each point
* [ ] update the rest of the data

lluiscab:
On my way to fix #200 

## TO DO List:

- [x] Fix indentation

#### Client
- [x] 0x04 login_plugin_request

#### Server
- [x] 0x02 login_plugin_response

#### Client
- [x] 0x11 declare_commands
- [x] 0x1d nbt_query_response
- [x] 0x31 face_player
- [x] 0x4c stop_sound
- [x] 0x54 declare_recipes
- [x] 0x55 tags

#### Server
- [x] 0x01 query_block_nbt
- [x] 0x0b edit_book
- [x] 0x0c query_entity_nbt
- [x] 0x15 pick_item
- [x] 0x1c name_item
- [x] 0x1f select_trade
- [x] 0x20 set_beacon_effect
- [x] 0x22 update_command_block
- [x] 0x23 update_command_block_minecart
- [x] 0x25 update_structure_block


lluiscab:
Okay, half an hour later I have the protocol.json updated, just need someone to take a look and make sure it's all ok.

AnthonyTsang:
To whom it may concern, simply updating this repo will not make mineflayer works. 1.13 replaces and changes some of the plugin channels that needs also update minecraft-protocol. Update details can be found [here](https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Plugin_Channels)

rom1504:
Mineflayer and node-minecraft-protocol don't really use plugin channels so that should be ok. But indeed there are changes to be made in mineflayer after this.

lluiscab:
I added all the packet Ids on my second commit.

AnthonyTsang:
Something still missing. I can only point out `smelting` in `packet_declare_recipes` at the moment.

lluiscab:
Package `packet_declare_recipes` was already added in 17w50a [here](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/17w50a/protocol.json#L3278)

AnthonyTsang:
You can read [the change](https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Declare_Recipes) 
17w50a's definition is missing `smelting`

lluiscab:
Just added the `smelting` definition. I'm not sure the `Ingredient` type is correct, just copied it from the other the `crafting_shaped` definition.

AnthonyTsang:
Sadly it still log out errors when connecting to my server. Will see if anything else is missing.

lluiscab:
Using mineflayer or nmp? As @rom1504 mentioned, changes will need to be made to mineflayer and probably to nmp too in order to support 1.13.

AnthonyTsang:
I have made some quick changes to make it run. You can follow this [list](https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Plugin_Channels) to change the plugin channel names (just search all and replace) in mineflayer and its minecraft-protocol. Also remember to update the minecraft-data of mineflayer to your version. In the list some of channels are dropped out and replaced with a dedicated packet, I dunno how to do it yet. Just handle the channel renaming works for me. After that, you should be able to at least login with mineflayer.

lluiscab:
I'll wait until this PR is merged, then try and make a simple client with NMP and fix whatever errors I might have, once NMP is 1.13 compatible, then mineflayer should be updated, that's at least how I think this should be done

AnthonyTsang:
Well I cant wait so long 😜 I dun know the best way, but the very initial test with the change shows there still some deserialization errors. 
There are problems with `entity_metadata`, `spawn_entity_living` and also I got a log `Chunk size is 2301 but only 2263 was read` which seems to relate to `packet_map` or `packet_map_chunk`
If you dun want to make the changes on mineflayer, you may want to compare those definition with the document.

lluiscab:
According to the wiki, `entity_metadata` and `spawn_entity_living` haven't changed.
I've updated `packet_map`.
''packet_map_chunk` has a property defined as buffer so it probably needs to be updated in code

AnthonyTsang:
Seems `entity_metadata` and  `spawn_entity_living ` is affected by the new metadata type  [here](https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Entity_Metadata_Format) they are adding two new types and it is defined at the beginning of the file.

AnthonyTsang:
The `particle` data type is used both in `entityMetadataItem` and `packet_world_particles`. The `packet_world_particles` is documented [here](https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Spawn_Particle)

AnthonyTsang:
Checkout `entityMetadataItem` for passing parameter to a type, you may need it when doing `particle`

lluiscab:
I'm updating `entityMetadataItem` right now, I'll push the changes later

lluiscab:
I've updated `entityMetadataItem` and `particle` but I'm not sure that It'll work properly, would you mind testing @AnthonyTsang?

data/pc/1.13/protocol.json
```diff
@@ -2016,6 +2016,18 @@
...
+                        "name": "direction",
+                        "type": "i8"
+					  },
+                      {
+                        "name": "hasDisplayName",
+                        "type": "bool"
+                      },
+                      {
+                        "name": "displayName",
+                        "type": ["option", "string"]
```
AnthonyTsang: `hasDisplayName` is duplicated and should be removed. The `option` is implying there is a boolean flag to either having the next data type or not.

AnthonyTsang:
The particle seems good but there are still error from `unlock_recipes` and `tags`. I am looking on it.

data/pc/1.13/protocol.json
lluiscab: `hasDisplayName` is a boolean value indicating whatever the map icon has a custom display name or not. See [here](https://wiki.vg/index.php?title=Pre-release_protocol#Map)

AnthonyTsang: You are right on the existence of this boolean, but `option` already imply there is such boolean. 
There are 2 equivalent ways to write such logic.
```
{
  "name": "displayName",
  "type": ["option", "string"]
}
```
or 
```
{
  "name": "hasDispalyName",
  "type": "bool"
},
{
  "name": "displayName",
  "type": [
    "switch",
    {
      "compareTo": "hasDisplayName",
      "fields": {"true": "string"},
      "default": "void"
    }
}
```

AnthonyTsang:
https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Unlock_Recipes
please add this according to the document to `packet_unlock_recipes`
```
            {
              "name": "smelting_recipe_book_open",
              "type": "bool"
            },
            {
              "name": "smelting_recipe_book_filter_active",
              "type": "bool"
            },
```

data/pc/1.13/protocol.json
lluiscab: I don't see why we should completely ignore a packet field just because it's redundant.

AnthonyTsang:
https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14150#Tags
And this to the end of `packet_tags`. Dun know why the document is not highlighting this field. 
```
            {
              "name": "fluidTags",
              "type": "tags"
            }
```

lluiscab:
Just pushed the changes, please check everything is correct now,

AnthonyTsang:
After `hasDisplayName`, `packet_tags`, and `packet_unlock_recipes` I dun have any error pop up.

data/pc/1.13/protocol.json
AnthonyTsang: nono, we did not ignore it.

lluiscab:
Perfect! Are you using NMP or mineflayer? Just to make me an diea of what I'm going to need to update for mineflayer to work.

data/pc/1.13/protocol.json
lluiscab: You said `hasDisplayName is duplicated and should be removed`. I could change the option to a switch but is that really necessary?

AnthonyTsang: Say we have a packet
```
01 02 03 04 05 
```
there are 5 bytes in total.
Assume boolean takes 1 byte and string takes 4 bytes.

```
{
  "name": "displayName",
  "type": ["option", "string"]
}
```
means the first byte is a boolean flag and the rest is a string. It correctly parses the packet and name the last 4 bytes to "displayName".

Whereas,
```
{
  "name": "hasDispalyName",
  "type": "bool"
},
{
  "name": "displayName",
  "type": ["option", "string"]
}
```
Means the first byte is a boolean flag and name it to `hasDispalyName`. The next byte is another boolean flag, and the rest of the bytes is a string but only 3 bytes left, so there is a mismatch.

lluiscab: Ooooh I see, I didn't know the `option` type checked for a boolean, okay, I'll update the `protocol.json` without the `hasDisplayName`

AnthonyTsang:
I am using mineflayer but since mineflayer is using minecraft-protocol (if it is what you mean by NMP) both of them needs to update. 
Maybe we should move to https://github.com/PrismarineJS/node-minecraft-protocol/issues/576 I have a comment there.

lluiscab:
I'll leave this PR waiting for approval then.

AnthonyTsang:
Ohohoh wait one last thing, `dataPaths.json` seems incorrect. I guess most of them should fallback to use 1.12? Take a look on `17w50a` in the same file.

lluiscab:
I guess so, but since I'm only updating the protocol, I'll leave that decision to @rom1504 or any other project maintainer.

rom1504:
Can you complete the files in pc/common with the new version ?
That will make circle ci run the tests on your protocol.json file as I think there might be some things wrong.

lluiscab:
Just updated the `pc/common` so the tests would run and as you said, there's an error with the `particleData`. I made `particleData` it's own type as it's quite large and appeared several times, I wasn't sure how to implement the `$compartTo` so I just looked on the file for somewhere similar and copied that. I'd need someone to explain me how to fix it properly.

AnthonyTsang:
I have a problem related to chunk data. Mineflayer blocks plugin `blockAt()` returns a wrong block at the correct location (the returning position is correct). It seems to me the chunk data is not parsed correctly despite no error comes out. Looking into it.

lluiscab:
@AnthonyTsang I think it might be better to open an issue on the mineflayer repo for that. As now blocks and items have different IDs it's very possible that some block ids have changed. Either way, that's not an issue for minecraft-data

AnthonyTsang:
For the particle type,

```

    "particle": [
      "container",
      [
        {
          "name": "particleId",
          "type": "varint"
        },
        {
          "name": "data",
          "type": [
            "particleData",
            {
              "compareTo": "particleId"
            }
          ]
        }
      ]
    ],
```

rom1504:
@AntonyTsang yes the way block are handled changed from 1.12 to 1.13 so it will require some work in mineflayer but indeed not a concern for the protocol.json
It might be one for block.json but that'll be for another pr.

@lluiscab nice that tests are passing.

I'll try and review it fully tonight.

lluiscab:
Any update?

rom1504:
Thank you for this PR.

We are still missing the other data here (blocks, items, ...) which can be extracted using the extractors, but let's do that in another PR. These data are required for mineflayer. However they are not required for node-minecraft-protocol so I will do a release of node-minecraft-data so we can start testing in nmp.

PR #204, opened 2018-08-25
Added more 1.13 data

lluiscab:
Added more 1.13 data. See #200 

- [x] blocks
- [x] biomes
- [x] effects
- [x] enchantments
- [x] entities
- [x] instruments
- [x] items
- [x] language
- [x] materials
- [x] protocol (Was already done before this PR)
- [x] recipes
- [x] windows


lluiscab:
Tests fail because I've added new values for pretty much everything. Possible block states, minecraft text ids (ex: "acacia_door")

lluiscab:
> However are you sure you have all the information included in 1.12 ?
What do you mean? 1.13 changed how block and items are stored so I'm pretty sure little to no data could be used from the 1.12 spec, could be wrong on that tho.

rom1504:
No I mean the 1.13 files should contain all the fields that were present in
1.12 files. No regression.

lluiscab:
See my PR on burger-extractor, I added some new 1.13 fields but the previous ones are still present

rom1504:
See
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12/blocks.json
Vs
https://github.com/lluiscab/minecraft-data/blob/154fbb10fd0722ebb74a5ae736c6f42debffb795/data/pc/1.13/blocks.json

Fields are missing as said by tests.

lluiscab:
Okay I see, I'll look into that tomorrow

lluiscab:
It's finally done, all data has been updated. Waiting for a review now @rom1504

rom1504:
Can you resolve the conflicts ? (git fetch && git rebase origin/master)

schemas/blocks_schema.json
```diff
@@ -76,53 +80,91 @@
...
-        }
+            "required": [
+              "name",
+              "type",
+              "num_values"
+            ],
+            "additionalProperties": false
+          }
+      },
+      "drops": {
```
rom1504: why did you change that ?

schemas/items_schema.json
```diff
@@ -12,6 +12,10 @@
         "type": "integer",
         "minimum": 0
       },
+      "textId": {
```
rom1504: isn't that the same as name ?

schemas/recipes_schema.json
```diff
@@ -2,11 +2,10 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "recipes",
   "description":"A dictionary of quoted numerical item IDs. Each ID maps to a list of recipes. There may be multiple different recipes per item (same ID and metadata). The recipes may not be sorted.",
-
   "definitions": {
     "id": {
-      "type":["integer","null"],
-      "description":"A single numerical ID or null."
+      "type":["string", "integer","null"],
```
rom1504: why not just continue using integer ids ?

schemas/recipes_schema.json
```diff
@@ -31,14 +30,26 @@
       },
       "required": ["id"],
       "additionalProperties":false
-    },
+	},
+	"textId_count":{
```
rom1504: why is that needed ?

schemas/blocks_schema.json
```diff
@@ -12,6 +12,10 @@
         "type": "integer",
         "minimum": 0
       },
+      "textId": {
```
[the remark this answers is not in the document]
lluiscab: It should be and it is for pretty much all the blocks that I've seen, but burger extracts them separately which means they're not the same for the game. So I guess we should keep both values

schemas/recipes_schema.json
lluiscab: The game doesn't anymore, why should we?

rom1504: yes it does ? the protocol still has integer ids everywhere.

schemas/blocks_schema.json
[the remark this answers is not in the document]
rom1504: "burger extracts them separately which means they're not the same for the game" are you sure ?

schemas/blocks_schema.json
```diff
@@ -27,7 +31,7 @@
           "number",
           "null"
         ],
-        "minimum": 0
+        "minimum": -1
```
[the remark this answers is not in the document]
lluiscab: Unbreakable blocks have a hardness of -1 according to the wiki (See [command block](https://minecraft.gamepedia.com/Command_Block) or the [block breaking trivia](https://minecraft.gamepedia.com/Breaking#Trivia))

schemas/blocks_schema.json
lluiscab: Change what? Not really sure what' you're talking about here

schemas/blocks_schema.json
```diff
@@ -146,7 +188,6 @@
       "displayName",
       "name",
       "hardness",
-      "stackSize",
```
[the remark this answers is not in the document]
lluiscab: I'll make a change to the burger-extractor in order to extract this value as well

schemas/recipes_schema.json
lluiscab: Okay, I get it. But for recipes, does it really matter?

schemas/recipes_schema.json
lluiscab: Metadata is no longer required, so a recipe output that just used a textid and count as params was needed

schemas/items_schema.json
lluiscab: It's probably the same that's happening with the block schema

schemas/blocks_schema.json
rom1504: The drop schema

schemas/items_schema.json
rom1504: Yes

schemas/blocks_schema.json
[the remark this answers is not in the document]
lluiscab: In 1.12, barrier hardness is set to `null`

schemas/recipes_schema.json
rom1504: Not sure what you mean. No it doesn't matter to use one or the other.
However it matters to have consistent files between versions or else it means having separate implementations for each version which is not great.
That doesn't seem too hard to convert recipes to integer ids ?

schemas/blocks_schema.json
lluiscab: I just made it possible for the drop array to contain strings for textIds, not really sure why github is saying I changed that much, maybe it's a spacing problem

schemas/recipes_schema.json
rom1504: Ok that will solve itself once that is switched back to integer ids.

schemas/recipes_schema.json
lluiscab: The game is now using text ids, and so was in 1.12, using text ids makes the files way more readable. Also the recipes don't use numeric ids anywhere in the vanilla game

rom1504:
Ok I will answer more in details to your point but just a general remark :
if we use the textid instead of the integers then we have to release major
versions of every prismarine package and clearly advertise that we are
breaking all the APIs ans everybody need to change their code if they want
to use 1.13

That's what I would like to try to avoid.

schemas/blocks_schema.json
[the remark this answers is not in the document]
rom1504: Yes Minecraft data name is what you called text id

schemas/blocks_schema.json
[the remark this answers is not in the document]
rom1504: Ok what about we continue using null instead of -1 ?

schemas/blocks_schema.json
rom1504: Ok that can be solved by using integer ids

schemas/recipes_schema.json
rom1504: Good point but really that file is not highly readable anyway so I'm not sure if we should change this...

schemas/blocks_schema.json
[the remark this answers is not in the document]
lluiscab: I have updated the burger extractor so it does not extract the textId anymore. I'll add the changes to this PR soon

schemas/blocks_schema.json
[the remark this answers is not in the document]
lluiscab: We could do that but why? Wouldn't it be better to have the value be always an integer?

schemas/blocks_schema.json
lluiscab: So block drops should go back to using the block id?

schemas/blocks_schema.json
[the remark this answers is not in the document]
rom1504: It's kind of the same thing so no real be benefit, however any change is a drawback because code using that need to be changed.

schemas/biomes_schema.json
```diff
@@ -12,6 +12,10 @@
         "type": "integer",
         "minimum": 0
       },
+      "textId": {
```
rom1504: That's the same as name isn't it ?

lluiscab: It probably is, will remove it if so.

schemas/items_schema.json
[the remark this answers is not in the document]
lluiscab: Yes, this value is still here, just need to get some time to remove it

schemas/blocks_schema.json
```diff
@@ -27,7 +27,7 @@
           "number",
           "null"
         ],
-        "minimum": 0
+        "minimum": -1
```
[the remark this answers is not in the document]
lluiscab: It's still there and I don't see why this shouldn't be changed to always a number, it's easier for type based coding and it's also cleaner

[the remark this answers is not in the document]
rom1504: in that case, it needs to be changed to all other versions and everything using that value (in mineflayer and flying-squid at least)

schemas/items_schema.json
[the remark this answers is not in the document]
rom1504: ok

schemas/blocks_schema.json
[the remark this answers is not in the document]
rom1504: it's not a question of whether it's better, but whether it's worth the effort to change it.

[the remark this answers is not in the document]
rom1504: I'd say if you want to change that, it would be simpler to do it in another PR or this is going to get complex

schemas/biomes_schema.json
lluiscab: For biomes, the name and textId are not the same, name is actually the display name of the biome (Ex: Frozen Ocean) and the textId is well, the id (ex: frozen_ocean)

schemas/items_schema.json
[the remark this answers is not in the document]
lluiscab: textId for items has been removed

schemas/blocks_schema.json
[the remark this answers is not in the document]
lluiscab: Fixed

schemas/biomes_schema.json
rom1504: Ok. Could you call textid name and what is currently name would become "displayName" ?
That won't break anything since we don't use biomes that much.

lluiscab: Yeah expect that changing the schema to add displayName will basically make all the biomes.json prior to 1.13 fail

rom1504: Just make it optional

lluiscab: Fixed!

schemas/recipes_schema.json
lluiscab: Fixed

schemas/recipes_schema.json
```diff
@@ -2,11 +2,10 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "recipes",
   "description":"A dictionary of quoted numerical item IDs. Each ID maps to a list of recipes. There may be multiple different recipes per item (same ID and metadata). The recipes may not be sorted.",
-
   "definitions": {
     "id": {
       "type":["integer","null"],
-      "description":"A single numerical ID or null."
+      "description":"A single numerical ID, text id or null."
```
rom1504: Remove that from the schema now that IDs are all int

schemas/recipes_schema.json
```diff
@@ -89,11 +88,10 @@
       ]
     }
   },
-
   "type": "object",
   "uniqueItems": true,
   "patternProperties" : {
-    "^[0-9]+$": {
+    "^[0-9 A-z]+$": {
```
rom1504: Remove that now that IDs are all int

schemas/recipes_schema.json
lluiscab: Done

schemas/recipes_schema.json
lluiscab: Done

lluiscab:
Everything should be ready

data/dataPaths.json
```diff
@@ -339,12 +339,12 @@
       "language":"pc/1.12"
     },
     "1.13": {
```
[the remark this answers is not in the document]
lluiscab: Will fix tomorrow, just a typo when I rebased

rom1504:
Things still missing : 
* in biomes : nothing
* in blocks : 
comparing https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12/blocks.json vs 
https://github.com/PrismarineJS/minecraft-data/blob/bdd4ff1c83680c8bf00aabce16657644a702ba6d/data/pc/1.13/blocks.json
  * [ ] harvestTools : sure it's an optional property, but it's very useful for bots to know what tool to use (+ it's available in the wiki)
  * [ ] variations : is that completely related by new blocks now ?
  * [ ] material : I believe that's only available in the wiki but this is required for the digging time computation algo which is required to dig a block in mineflayer
* in effects : nothing
* in enchantments : nothing
* in entities : nothing
* in instruments : nothing
* in items : nothing
* in language : nothing
* [ ] material.json is missing
* in protocol : nothing
* in recipes : looks ok but honestly it's hard to be sure
* in version : nothing
* in windows : nothing

Things that really should be fixed : 
* [ ] point all files to 1.13 in datapath
* [ ] copy material.json
* [ ] material property in blocks.json

Optional but would be nice
* [ ] harvestTools : not used in mineflayer, but very useful for bots

lluiscab:
Things that should be fixed:

- [x] point all files to 1.13 in datapath
- [x] copy material.json
- [ ] material property in blocks.json
- [ ] harvestTools property in blocks.json

rom1504:
the harvestTools value and the content of materials.json are wrong.
Reading the burger extractor code, I'm under the impression it is just copying the old (1.12) data if it cannot find it.
I can't see how that can produce up to date data.

lluiscab:
Yikes for the id change, will see if I can update that tomorrow

lluiscab:
Should be all ready now (Hope this time is true)

rom1504:
looks good, merged, thanks for all the work on this !

PR #208, opened 2018-09-28
Fix incorrectly named field in 1.13 packet_face_player

nickelpro:
My compiler was complaining about a duplicated field for this packet.

The final double is labeled `x` when it should be `z`.

This patch fixes that, also a bunch of random trailing whitespace my editor autoremoved.


rom1504:
good catch

PR #209, opened 2018-09-30
Removed redundant metadata from 1.13 recipes

timmyrs:


rom1504:
if you want to do this, change it in the extractor script and then apply it.
However I can't really see the point of removing that information.

PR #226, opened 2019-02-20
Fix armor being stackable (fixes #223)

### LINKED ISSUE #223: armor is not stackable
The current 1.13.2 item lists helmets (and others) as 64-stack items:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.13.2/items.json#L2826
However, those don't stack in-game.

timmyrs:


rom1504:
Let's merge.
Be aware these changes might be overwritten if we rerun the script generating the data though.
A better solution would be to put the manual data in an other file somewhere.

PR #228, opened 2019-03-05
Fix block names

timmyrs:


rom1504:
Source ?

timmyrs:
Thanks for asking, I thought it was always called grass_block, but apparently not.

PR #232, opened 2019-04-24
Fix advancements packet (u64 -> i64) (fixes #217)

### LINKED ISSUE #217: Native type for u64 isn't imported for 1.12.2 protocol.json
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12.2/protocol.json#L979
There is no "u64": "native" in any of parent namespace types

timmyrs:
An unsigned long (u64) is never used in Minecraft, and the advancements packet criterion progress uses a normal long (i64) according to wiki.vg.

### DIFF
--- a/data/pc/1.12-pre4/protocol.json
+++ b/data/pc/1.12-pre4/protocol.json
@@ -962,7 +962,7 @@
                                   "name": "criterionProgress",
                                   "type": [
                                     "option",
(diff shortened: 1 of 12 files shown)


rom1504:
it makes no sense (as a design decision from mojang), but indeed you're right

PR #233, opened 2019-04-25
Add some initial data for 1.14

timmyrs:
Burger data extracted using https://gist.github.com/timmyrs/7ff2fdf8c8acee895f5a7fcdfaa3cb32


rom1504:
Could you use https://github.com/PrismarineJS/burger-extractor (and fix it if needed) ?

timmyrs:
I can submit the fixes I applied to items.json to burger-extractor but otherwise I think my PHP script output the same data as burger-extractor.

timmyrs:
Alright, I've patched up burger-extractor, ran it, and used its data, instead. However, the blocks.json it generated seems to be faulty: https://circleci.com/gh/PrismarineJS/minecraft-data/692

data/pc/common/protocolVersions.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
+    "minecraftVersion": "1.14",
+    "version": 477,
+    "usesNetty": true,
+    "majorVersion": "1.14"
+  },
   {
     "minecraftVersion": "1.13.2",
```
rom1504: We need all versions in that file

rom1504:
Let's merge as a beginning.
We will need to run the more complete generator (burger-extractor) next. Fixing it also means making it produce valid output.

PR #237, opened 2019-05-18
1.14.1 support

rom1504:
using https://wiki.vg/index.php?title=Pre-release_protocol&oldid=14764


timmyrs:
I think there were no protocol changes between 1.14 and 1.14.1 so that protocol.json can be duplicated/linked to 1.14 as well.

rom1504:
true

rom1504:
actually there may have been changes between 1.14 and 1.14.1, I'm not linking both protocols for now https://wiki.vg/index.php?title=Pre-release_protocol&type=revision&diff=14764&oldid=14701

PR #244, opened 2019-06-23
[docs] bump minecraft-data to 2.37.5

RONTheCookie:
didn't test but according to the rules of semver it should be compatiable


RONTheCookie:
whoops

rom1504:
Won't do anything.
The commit need to go into web branch and then need to run npm run publish.
I'll do it

PR #249, opened 2019-07-16
Fix protocol version of ~1.9.4 and add missing versions

edomora97:
Fixed some wrong protocol versions for minecraft 1.9.3-pre3 till 1.9.4 (was 109 instead of 110).
Checking https://wiki.vg/Protocol_version_numbers I've also added some missing versions to the file.


rom1504:
Looks good, thank you for the PR

PR #250, opened 2019-08-08
Data Corrections for blocks

kamesa7:
I edited mainly emitLight, displayName, grass material and instruments.
I deleted variations with normal glass and glass_pane.


rom1504:
@timmyRS could you check how that looks ?

timmyRS:
All changes look correct to me; I cross-referenced with Minecraft Wiki.

Issue #240, opened 2019-06-04
Why pc/common/versions.json?


timmyRS:
I understand that pc/common/protocolVersions.json contains all protocol versions, regardless of inclusion in minecraft-data. However, pc/common/versions.json only contains versions included in minecraft-data. Isn't this pointless as the same data can be obtained by taking all keys in dataPaths.json's "pc" object?

rom1504:
Yeah we could do that indeed (use the keys of loadpath and remove common/versions)

PR #259, opened 2019-12-19
Add data versions

timmyRS:
I know that "dataVersion" is never actually used in the protocol, so it might not be a perfect fit for "protocolVersions.json" but since "versions.json" is already taken (related: #240), this is the best solution to remain backward-compatibility.

### DIFF
--- a/data/pc/common/protocolVersions.json
+++ b/data/pc/common/protocolVersions.json
@@ -2,1260 +2,1477 @@
   {
     "minecraftVersion": "1.15.1",
     "version": 575,
+    "dataVersion": 2227,
     "usesNetty": true,
     "majorVersion": "1.15"
   },
   {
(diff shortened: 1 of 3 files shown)


rom1504:
What is this data used for ? Is it coming from Mojang ?

timmyRS:
It's used for worlds and structures to determine which version they were made in/for, and I'd use it for reading them.

The data is sourced from https://minecraft.gamepedia.com/Protocol_version

rom1504:
Can you still run your automatic updater if you add this data ?

timmyRS:
Yes, this data is also included in the client.jar's version.json.

rom1504:
ok cool

timmyRS:
They actually renamed the field to `world_version` in the version.json now :roll_eyes:

PR #271, opened 2020-04-18
Add category property back

ImHarvol:
In https://github.com/PrismarineJS/minecraft-data/commit/7e0b470872d7e3d1de937d7099dd547fe62e0a2b Entity.category was removed from 1.13, which was pretty useful to distinguish between hostile and non-hostile mobs (https://github.com/PrismarineJS/mineflayer/pull/941#discussion_r410127163).

I added them back using data from https://github.com/PrismarineJS/minecraft-wiki-extractor but I didn't get the category of some entities that were added in the same commin, so I'm not sure if that could be a problem.

### DIFF
--- a/data/pc/1.13.2/entities.json
+++ b/data/pc/1.13.2/entities.json
@@ -6,7 +6,8 @@
     "displayName": "AreaEffectCloud",
     "width": null,
     "height": 0.5,
-    "type": "mob"
+    "type": "mob",
+    "category": "Immobile"
   },
   {
     "id": 1,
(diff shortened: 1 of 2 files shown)


rom1504:
which page and revision from minecraft wiki did you use for this ?

ImHarvol:
> which page and revision from minecraft wiki did you use for this ?

I just used `node entities_to_json.js  entities.json 2020-01-01T00:00:00Z` and then copied the entities.json's .category of each entity to the entities.json in the minecraft-data repo.
I dont know If I explained myself hehe

PR #273, opened 2020-04-29
Add 1.14.4 biomes blocks entities items and recipes data

Karang:
1.14.4 data generated with Burger and burger-extractor


rom1504:
Cool.
You need to change the datapath.json file too for this to work

rom1504:
Did you have to adapt stuff in burger extractor ?
If you did could you also pr these changes ?

ImHarvol:
Every mob has _"Hostile mobs"_ as his category

Karang:
No, I didn't change burger extractor. The only modification I did were manual (2nd commit).
The goal is to provide a base on which to improve, I'm sure there are many mistakes, but having this could already unlock the works on the other packages (prismarine-chunk, mineflayer...)

rom1504:
thanks, released in mcdata 2.47.0

PR #276, opened 2020-05-03
Fixed functionality-breaking typo

Plazmaz:
This JSON is invalid and breaks the whole library.


rom1504:
Tests should have been failing.
Something is wrong

PR #279, opened 2020-05-03
Typo in (at least 1.14.4's) blocks.json

Shana6:
Just a quick change, someone should check other versions for this typo.


rom1504:
this is already fixed, I don't understand why github still display a diff

Shana6:
weird, thanks for the info lol

PR #280, opened 2020-05-04
Fix elytra handling, add optvarints

Mstrodl:
For some reason the OptVarint type of `entity_metadata` (type 17) was set to `["option", "varint"]` which is incorrect, I've also made a PR for `node-minecraft-protocol` to add support for the new optvarints. See [this bug](https://bugs.mojang.com/browse/MC-111480) for the reasoning behind this

### DIFF
--- a/data/pc/1.14.1/protocol.json
+++ b/data/pc/1.14.1/protocol.json
@@ -1,6 +1,7 @@
 {
   "types": {
     "varint": "native",
+    "optvarint": "native",
     "pstring": "native",
     "u16": "native",
     "u8": "native",
@@ -228,7 +229,7 @@
               }
             ]
           ],
-          "17": ["option", "varint"],
+          "17": "optvarint",
           "18": "varint"
         }
       }
--- a/data/pc/1.14.1/protocol.json
+++ b/data/pc/1.14.1/protocol.json
@@ -1,7 +1,7 @@
 {
   "types": {
     "varint": "native",
-    "optvarint": "native",
+    "optvarint": "varint",
     "pstring": "native",
     "u16": "native",
     "u8": "native",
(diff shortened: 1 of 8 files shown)


rom1504:
Is there any drawback to just using varint ?

Introducing a new native type for this weird format doesn't sound like a good idea. Entity metadata still need a bunch of handling after the node-minecraft-protocol parsing anyway

Mstrodl:
@rom1504 That's what I did initially, but I figured that it might make more sense to just do it the proper way (I figured I would get yelled at for doing it the cheat way to be entirely honest). I figured it would make the most sense to just do it in the way that would be most consistent.

rom1504:
consistent with what ?
mojang decided to make 0 a special value, that's their choice, I think we should probably be consistent with that.
I don't think we should be trying to "fix" the protocol at this layer, or we would have lot of work that wouldn't make that much sense.

So actually I think the varint solution is ok.

I'm particularly reluctant to adding a new type for this case as one of the objective of these protocol definition is to rely as little as possible on native implementation of datatypes and to have as generic as possible datatypes (to make it easier to reimplement in js or other languages)

Mstrodl:
@rom1504 That's fair enough, but I think in that from the user's point of view, wouldn't the value be expected to decode the same way as the rest of the entity ids? We aren't fixing Mojang's protocol, we're just decoding it the same way they do. Saying it's a varint would be technically true, but it's not entirely accurate. Is it perhaps possible to at least declare `optvarint` as `varint` instead of `native`? That way the docs show it as different at least

rom1504:
yeah you can do that (just put varint where you put native)

Mstrodl:
@rom1504 Alright, should I close my PR for node-minecraft-protocol then? Or will protodef still let me declare a handler for optvarint even though it isn't `native`?

rom1504:
you can just close it then since there's nothing more to be done as optvarint is just a varint now

PR #281, opened 2020-05-05
Fix digging in mineflayer 1.14.4

Karang:
@rom1504 This should fix the digAndBuild test for 1.14.4


rom1504:
Could you copy to file to the 1.14.4 folder instead ?

Karang:
> Could you copy to file to the 1.14.4 folder instead ?

done

rom1504:
yeah but actually no it's wrong, these item ids changed in 1.14
I'm working on updating this

rom1504:
done in #282

PR #287, opened 2020-05-20
add enchantments from wiki

uncovery:
generated with this code https://github.com/PrismarineJS/minecraft-wiki-extractor/pull/27

### DIFF
--- a/data/pc/1.15.2/enchantments.json
+++ b/data/pc/1.15.2/enchantments.json
@@ -1,0 +1,746 @@
+[
+  {
+    "itemName": "minecraft.aqua_affinity",
+    "displayName": "Aqua Affinity",
+    "maxLevel": 1,
+    "primaries": [
+      "minecraft.golden_helmet",
+      "minecraft.iron_helmet",
+      "minecraft.leather_helmet",
+      "minecraft.chainmail_helmet",
+      "minecraft.turtle_helmet"
+    ],
+    "secondaries": [],
+    "excludes": [],
+    "probability": 2
+  },
+  {
+    "itemName": "minecraft.bane_of_arthropods",
+    "displayName": "Bane of Arthropods",
+    "maxLevel": 5,
+    "primaries": [
+      "minecraft.golden_sword",
+      "minecraft.iron_sword",
+      "minecraft.stone_sword"
+    ],
+    "secondaries": [
+      "minecraft.golden_axe",
+      "minecraft.iron_axe",
+      "minecraft.stone_axe"
+    ],
+    "excludes": [
+      "minecraft.smite",
+      "minecraft.sharpness"
+    ],
+    "probability": 5
+  },
+  {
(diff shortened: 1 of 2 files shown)


uncovery:
Seems there is a required format for enchantments that I was not aware of? 
Need to find out where that is defined and fix it.

rom1504:
Mineflayer is using that schema for example. If you can add property without changing current properties it's good.

uncovery:
Ok, sounds reasonable. Question:

how can we make sure the unique integer ID is consistent? This ID has been deprecated and now it's the name string (e.g. minecraft:efficiency).

rom1504:
I can't find any place in prismarinejs where we use the enchantments.
Do you have an use case in mind ?
What packets would this be used with for example ?

Karang:
> I can't find any place in prismarinejs where we use the enchantments.
> Do you have an use case in mind ?
> What packets would this be used with for example ?

It would be useful to have that in mineflayer because enchantments can affect some properties:
* dig time is affected by the tool enchantment
* fall damages is affected by Protection and Feather falling enchantments
Also having this data could be useful to make the bot work with enchantment tables ?

rom1504:
Yeah, so same question then, if we want to use them in client/server, which packet is it ?
That'll make it possible to answer what data we need

rom1504:
Ok so the answer is for example https://wiki.vg/Protocol#Window_Property

As you can see the numerical IDs are still used and not deprecated at all.

So we do need them.

The wikis removing them means end users may not need them, so I guess you'd need to find these IDs elsewhere. (The jar I guess ? Have a look at burger extractor maybe)

uncovery:
> Ok so the answer is for example https://wiki.vg/Protocol#Window_Property

Ok, did not know that. Will take a look how to get them then.

uncovery:
> Do you have an use case in mind ?

My use case for example is that I can now make sure that a system that dishes out random enchanted items does create any that violate the enchantment conflicts.

uncovery:
> I can't find any place in prismarinejs where we use the enchantments.

I am a bit confused why this question is being asked. While I can understand that you guys want to have the enchantments be done properly (with the IDs and fitting here to the data schema etc, I would like to know what other prerequisites this needs - beyond just working. I have started doing this here instead of in my own systems mainly because there was already an older version doing exactly that, I am merely updating it. 

If there is a requirement to have an active use in prsimarineJS or some such, then I'd rather drop the ball here and re-do it in my own system instead. I have fulfilled my own requirements on this functionality a week ago and everything further from now on is done 100% to satisfy other people's needs. In case there is no such need, please let me know ASAP so I can focus my time on more productive things.

I know I am poor at programming and sorry if my mistakes and shortcomings are creating additional work. If it's not even achieving anything once it's done, this does not make any sense at all. Sorry if this sounds like a rant, I do not mean to be guilt-tripping people into liking what I do. If you do not see a use case, just tell me and I'll move on without hard feelings.

Issue #134, opened 2016-05-08
Fix nether block bounding box in 1.8


rom1504:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/1.8/blocks.json#L2325

the bounding box should be empty even though the mcwiki doesn't have the info

have a overwrite_blocks.json during the extraction

PR #288, opened 2020-05-27
Audit shapes and add missing shapes

Karang:
This PR adds a script to audit the block shapes.

The tests performed are:
* Bijection between blocks.json and blockCollisionShapes.json (every block should have a corresponding shape and every shape should have a corresponding block)
* Bijection between shapes.blocks values and shapes.shapes (every shape id should have a corresponding bounding box and every bounding box should be used by at least one block)
* Consistency between blockCollisionShapes.json  and blocks.json 's `boundingBox` property ("empty" if shape=0, else "block")

These tests allowed me to fix some errors in the existing shape files and to recreate manually the shape files for the missing version. (1.13 created from 1.14 shapes, 1.7-1.11 created from 1.12 shapes, by removing or renaming blocks: I assumed the bounding boxes didn't change)


Karang:
I added a consistency test with the blocks.json 's `boundingBox` property. I think it would be useful to fix the data so it is consistent with the new blockShapes. But it raises a few questions:
* do we want to continue using the boundingBox property or de we move toward using blockShape or a derived value ?
* at the moment `boundingBox` is either `empty` or `block`, what should we label the blocks that are neither ?

Karang:
I used the infos in blockCollisionShapes.json to fix automatically the `boundingBox` property.
It should closes #134 and improve current physic for versions > 1.13

rom1504:
Looks pretty good.

About bounding box property, let's keep it for now, with empty and block only.
It's ok not to use it in mineflayer and flying-squid and instead to use the block shapes.

data/pc/1.12/blockCollisionShapes.json
```diff
@@ -238,7 +238,7 @@
 "lime_glazed_terracotta":1,
 "pink_glazed_terracotta":1,
 "gray_glazed_terracotta":1,
-"silver_glazed_terracotta":1,
+"light_gray_glazed_terracotta":1,
```
Gjum, line 241: are you sure about this? they show up as silver_glazed_terracotta/silver_shulker_box in-game.

Karang: I updated it to be consistent with blocks.json https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12/blocks.json#L5792 It is also listed as light gray on the wiki https://minecraft.gamepedia.com/Terracotta

PR #291, opened 2020-06-05
Change ladder's bounding box to empty

IdanHo:
Players can walk through blocks in which ladders are placed, so i dont think a "block" boundingBox is appropriate

### DIFF
--- a/data/pc/1.10/blocks.json
+++ b/data/pc/1.10/blocks.json
@@ -1727,7 +1727,7 @@
     "hardness": 0.4,
     "stackSize": 64,
     "diggable": true,
-    "boundingBox": "block",
+    "boundingBox": "empty",
     "drops": [
       {
         "drop": 65
--- a/data/pc/1.11/blocks.json
+++ b/data/pc/1.11/blocks.json
@@ -1701,7 +1701,7 @@
     "hardness": 0.4,
     "stackSize": 64,
     "diggable": true,
-    "boundingBox": "block",
+    "boundingBox": "empty",
     "drops": [
       {
         "drop": 65
--- a/data/pc/1.12/blocks.json
+++ b/data/pc/1.12/blocks.json
@@ -1701,7 +1701,7 @@
     "hardness": 0.4,
     "stackSize": 64,
     "diggable": true,
-    "boundingBox": "block",
+    "boundingBox": "empty",
     "drops": [
       {
         "drop": 65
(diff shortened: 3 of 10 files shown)


Karang:
> Players can walk through blocks in which ladders are placed, so i dont think a "block" boundignBox is appropriate

It depends on the direction a player walk in. Ladders do have a solid bounding box (even if it is very thin). As a result, players can collide and even stand on the top of the ladder. It's similar to block like slab or carpets, the player can stand in the block but collide with some part of it.

The audit_shape.js tool is there to check consistency between the boundingBox property and the shapes array.  https://github.com/PrismarineJS/minecraft-data/blob/master/tools/js/test/audit_shapes.js

The convention is that a block is "empty" if the shape array length === 0.

IdanHo:
> Ladders do have a solid bounding box (even if it is very thin).
This is my main problem with it being assigned the "block" bounding box, since its so thin it cant be used as a "safe" platform block, ill see about checking the blockCollisionShapes instead, but adding better options to the boundingBox property like you suggested might be a good idea

Karang:
> > Ladders do have a solid bounding box (even if it is very thin).
> 
> This is my main problem with it being assigned the "block" bounding box, since its so thin it cant be used as a "safe" platform block, ill see about checking the blockCollisionShapes instead, but adding better options to the boundingBox property like you suggested might be a good idea

The question was also asked here https://github.com/PrismarineJS/minecraft-data/pull/288#issuecomment-634796191 I suggest we open an issue to talk about it and decide if we want to keep the boundingBox property and modify it to have more options or move completely to use the shapes array.

The question is also related to creating a prismarine-physic project to centralize the physic engine but also some utility functions to work with collisions.

To test the collision of a player with the environment, there is the adjustPositionHeight function https://github.com/PrismarineJS/mineflayer/blob/master/lib/physics.js#L85 Maybe it can be useful to check if the player can stand at a certain pos (it modify the y of the given pos assuming only gravity).

IdanHo:
btw, it seems like vines do have a boundingBox of 'empty', although they behave the exact same way as ladders when a block is placed behind them

IdanHo:
you are able to stand on top of them while pressing shift, though

Karang:
> you are able to stand on top of them while pressing shift, though

not only on top, you can stand anywhere in the block if you press shift. But on ladder you can stand on top without shift if your hitbox collides with the 3 pixel wide hitbox of the ladder (that's the difference with vines). It's the same hitbox as open trapdoors

IdanHo:
ah i see, so yeah, this PR can probably be closed, as its the most general issue that needs to be addressed.

PR #293, opened 2020-06-08
Add missing block items

Karang:
Before 1.13 (the flattening), all blocks had a corresponding item with the same id (some of them were unobtainable, but the item existed). In mineflayer, when we wanted to get the item corresponding to a block, we had to query the id in blocksByName instead of itemsByName.

With the flattening, things became more complicated as an item could have a different id from its block. So to get a blockitem id, we now have to use itemsByName. This lead to complicated code like this, everywhere in mineflayer :

```js
  let grassName
  let itemsByName

  if (bot.supportFeature('itemsAreNotBlocks')) {
    grassName = 'grass_block'
    itemsByName = 'itemsByName'
  } else if (bot.supportFeature('itemsAreAlsoBlocks')) {
    grassName = 'grass'
    itemsByName = 'blocksByName'
  }
```

This PR adds all the missing items from blocks for versions <1.13 so we are able to use itemsByName for blocks in whatever version we use (no more specialized code).

The items are generated using the corresponding blocks.json and the script audit_blockitems.js

### DIFF
--- a/data/pc/1.10/items.json
+++ b/data/pc/1.10/items.json
@@ -1,5 +1,1193 @@
 [
   {
+    "id": 0,
+    "displayName": "Air",
+    "name": "air",
+    "stackSize": 64
+  },
+  {
+    "id": 1,
+    "displayName": "Stone",
+    "name": "stone",
+    "stackSize": 64
+  },
+  {
+    "id": 2,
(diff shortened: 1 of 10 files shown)


rom1504:
good idea, I like it

PR #297, opened 2020-06-28
1.16.1 Blocks and more

AppDevMichael:


rom1504:
Thanks, looks good.
For reference, did you just run the burger extractor or did you do something else ?

rom1504:
we really need a script to get everything at once though, this is too much manual effort

AppDevMichael:
Yeah, I used the burger extractor but I had to edit it. I might try and write some documentation on how to use it and how to configure it.

Issue #298, opened 2020-06-28
1.16 data


rom1504:
missing : 
* [x] material.json, similar to https://github.com/PrismarineJS/minecraft-data/pull/282
* [x] language.json : minecraft jar extractor has a script
* [x] collision shape : see gjum mod

cc @AppDevMichael if you're interested

BTW a script doing everything for a new version would be awesome https://github.com/PrismarineJS/minecraft-data/issues/283

AppDevMichael:
#300 

Added all of the missing files. 

I will look into the scripting at some point.

rom1504:
Thanks, done

PR #299, opened 2020-06-29
Update history.md

AppDevMichael:


rom1504:
I think we should do https://github.com/PrismarineJS/minecraft-data/issues/298 before releasing
Without this data, we cannot use this in mineflayer

AppDevMichael:
I will try and work on it tomorrow it is 2am for me at the moment and I need sleep.

AppDevMichael:
Delete this

PR #301, opened 2020-06-29
Update block states to keep data generator order, add defaultState property

Karang:


tools/js/tools/patch_states.js
```diff
@@ -0,0 +1,72 @@
+const fs = require('fs')
```
rom1504, line 1: could you put this in https://github.com/PrismarineJS/minecraft-jar-extractor ?
I prefer to keep the amount of js in this repo to a minimum (just validation)

Karang: done https://github.com/PrismarineJS/minecraft-jar-extractor/pull/19

Issue #295, opened 2020-06-22
food.json


rom1504:
https://minecraft.gamepedia.com/Food parsing this and putting it in a food.json

Would make it easier to code bot that can eat

AppDevMichael:
What would one do for when there are two Suspicious Stews, how could I differentiate them in the json?

Current format that I have come up with:
```js
  {
    id: 576,
    displayName: 'Apple',
    name: 'apple',
    stackSize: 64,
    foodPoints: 4,
    saturation: 2.4,
    effectiveQuality: 6.4,
    saturationRatio: 0.6
  }
```

rom1504:
There is only one Suspicious stews. Apparently there is a special effect to take into account with it, but I'd say it doesn't belong in that file

AppDevMichael:
What about the cake?

rom1504:
well I guess you can always decide to put 2 names for them, and we'd have to be careful about that when using it

rom1504:
last step https://github.com/PrismarineJS/minecraft-data/pull/302#issuecomment-653774254

AppDevMichael:
https://github.com/PrismarineJS/node-minecraft-data/pull/37

rom1504:
done

PR #302, opened 2020-06-30
Adding foods

AppDevMichael:
#295 

Added foods


rom1504:
How do you plan to use it ? If for mineflayer, it would be great to try to extract the same for other versions. You can change the revision in wiki extractor.
Also could you commit your extractor in wiki extractor repo and link it here ?

AppDevMichael:
I there a list of dates or do I have to find the dates for each release?

rom1504:
you can probably find the dates in the wiki

Karang:
> I there a list of dates or do I have to find the dates for each release?

You can probably generate the older versions from the latest ones, by copying data + checking if the item existed (in items.json).
I don't think any food item were removed (only added). And at least for versions >= 1.9 there shouldn't be any change in properties.

AppDevMichael:
https://github.com/PrismarineJS/minecraft-wiki-extractor/pull/33

AppDevMichael:
I'll do it tomorrow

rom1504:
thanks, the next step so this is usable in mineflayer/flying-squid/node is adding the functions to read this in https://github.com/PrismarineJS/node-minecraft-data ; could you do it ?

AppDevMichael:
I will have a look into it, I may not have as much time to do so because of work.

rom1504:
ah items.json of 1.16.1 was apparently removed in this PR and apparently tests didn't catch it

PR #303, opened 2020-07-03
Get rid of a dead link

u9g:
Was an example, but I guess the repo was deleted, so I got rid of it from the readme


rom1504:
sadly it's gone indeed...

PR #306, opened 2020-07-16
Languages

redcarti:


rom1504:
How big is that ?
If it's more than 5MB total we need to think of another solution.
For example putting this in minecraft-assets might be better if that's the case

redcarti:
1.16.1 langs is 42mb total

Karang:
The problem with putting that in mc-assets is that in order to use the languages, we'll have to load all assets, which is a lot more than we need for the case of a server. Maybe have minecraft-language package ?

redcarti:
Yeah, I think that could be better

rom1504:
For sure in the JavaScript integration of this we want to load only the language we need at run time. So it will not affect ram usage.
However 42 MB for a package is a lot, especially one we use everywhere (minecraft data)

I believe the only use case for this data will be in prismarine-chat package (used in flying-squid and mineflayer)

There are several ways to bundle this :
* In minecraft asset npm package
* In a minecraft language npm package
* In a minecraft language *repo* that would not be a dependency for prismarine-chat but instead be something that is downloaded at demand in the package.

Do you see any drawback with the 3rd solution ? This is basically the way the vanilla launcher works (except we'd be using GitHub http server instead of our own), and a decent way to distribute assets for websites and games

redcarti:
hmm, in case we can use dynamic loading, e.g. `mcLang.en_US.string` and don't load all langs.

rom1504:
There is also a fourth solution which might be even better : directly use https://github.com/PrismarineJS/node-minecraft-wrap/blob/master/lib/launcher_download.js 
As done in https://github.com/PrismarineJS/minecraft-jar-extractor/pull/23/files

Directly in a npm package prismarine-language at run time

That way we don't store anything and download from mojang what we need.

Does that make sense ?

@redcarti could you tell us how much time it takes to retrieve 1 language for 1 version ? (Using this extractor)

We could still store only the en language in minecraft data as fallback

redcarti:
> could you tell us how much time it takes to retrieve 1 language for 1 version ? (Using this extractor)
4-5 sec

redcarti:
> We could still store only the en language in minecraft data as fallback

I didn't deleted language.json file so, yeah, I think we could

rom1504:
Ok that sounds like a good solution then, could you try to use this extractor directly in flying-squid at run time @redcarti to see if it works well for you ?

If that works well we'll move that part to a package that exposes a good way to access minecraft client assets (or maybe directly use minecraft-wrap everywhere if that's convenient)

redcarti:
eh, where do I need to take mc files? I have `path.join(process.env.APPDATA, '/.minecraft')` in lang extractor

redcarti:
leave it as it is?

rom1504:
yes you can leave it as is

FYI this is not the path where it takes the assets but the path where it *saves* them.
It automatically downloads the files from mojang

AppDevMichael:
Why not create a new repo to include all of the langs and then a different branch that is "lighter" so they only have English for example

rom1504:
we went a different way which is to extract language are run time, I believe @redcarti has started it at https://github.com/PrismarineJS/flying-squid/pull/414

AppDevMichael:
That pull request was created before this one was so I don't think that is true. It would be nice to have outside of flying squid either way for example, creating a wiki of some sort.

AppDevMichael:
nvm

rom1504:
It is true, we talked.
Yes sure it'll get extracted out of flying squid whenever it works. (The function that gets out any language for any version)

redcarti:
Close?

rom1504:
Yeah, stale

PR #311, opened 2020-07-20
Added correct drop data for blocks (1.13+)

TheDudeFromCI:
The drop data for blocks in Minecraft versions 1.13+ was inaccurate due to conversion errors after the flattening. This commit fixes the block drop data by pulling information directly from the Minecraft data folder within the jar.


rom1504:
Great !
Could you put the extraction script in minecraft jar extractor ?

rom1504:
is this code https://github.com/PrismarineJS/minecraft-jar-extractor/pull/24/files generating these blocks.json files ? it doesn't seem to be doing that

rom1504:
Could you add a mention of this extractor in the readme ?
The procedure to extract everything for a new version is becoming more and more complicated. Would be great if we had a script that would do it end to end

rom1504:
Jar extractor I guess but I think it would be good to write a small guide
to extract all the data of minecraft data.
Maybe you can start it with just what you added here ? (Put it in docs/ )

PR #313, opened 2020-07-26
Fix "feet_eyes" field of packet_face_player

nickelpro:
"feet_eyes" is wrong for packet_face_player, it's a varint encoded enum according to [wiki.vg](https://wiki.vg/Protocol#Face_Player) as well as Burger:

```
"PLAY_CLIENTBOUND_34": {
    "class": "pj.class",
    "direction": "CLIENTBOUND",
    "from_client": false,
    "from_server": true,
    "id": 52,
    "instructions": [
        {
            "field": "e",
            "operation": "write",
            "type": "enum"
        },
        {
            "field": "a",
            "operation": "write",
            "type": "double"
        },
        {
            "field": "b",
            "operation": "write",
            "type": "double"
        },
        {
            "field": "c",
            "operation": "write",
            "type": "double"
        },
        {
            "field": "g",
            "operation": "write",
            "type": "boolean"
        },
        {
            "condition": "g",
            "instructions": [
                {
                    "field": "d",
                    "operation": "write",
                    "type": "varint"
                },
                {
                    "field": "f",
                    "operation": "write",
                    "type": "enum"
                }
            ],
            "operation": "if"
        }
    ],
    "state": "PLAY"
},
```

Checked every version with Burger, the only thing that changes is the packet id, it's always been a varint.

### DIFF
--- a/data/pc/1.13.1/protocol.json
+++ b/data/pc/1.13.1/protocol.json
@@ -1518,7 +1518,7 @@
             [
               {
                 "name": "feet_eyes",
-                "type": "string"
+                "type": "varint"
               },
               {
                 "name": "x",
--- a/data/pc/1.13.2-pre1/protocol.json
+++ b/data/pc/1.13.2-pre1/protocol.json
@@ -1523,7 +1523,7 @@
           [
             {
               "name": "feet_eyes",
-              "type": "string"
+              "type": "varint"
             },
             {
               "name": "x",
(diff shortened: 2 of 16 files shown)


rom1504:
Thanks, I guess the server rarely send this packet for us to have never seen this issue.
We have a plan (almost there actually) to automatically dump bunch of packets from minecraft to be able to tests node-minecraft-protocol (but also maybe other implementations of the protocol ?)

Issue #316, opened 2020-07-28
Changes in Particle IDs need to be tracked, breaks packet_world_particle


nickelpro:
[Particle IDs](https://wiki.vg/Protocol#Particle) change between MC versions, and minecraft-data should have a way to track that.

Besides semantic differences between particle IDs, the protocol is broken for several versions right now due to particle data using [hard coded values](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.1/protocol.json#L91-L143) that have changed due to the addition of new particle IDs (in this case, Dripping Lava, Landing Lava, and Falling Water).

My recommendation is to create a particles.json to track the particle ids and reference these in the particleData type or something similar.

rom1504:
interesting, thanks for noticing this
I'm curious how you find these problems with the protocol definition.
Any particular method or you just compare manually?

nickelpro:
I'm writing a C++ protocol generator based on minecraft-data. Every time I implement a new type I verify it with a minimal server I've implemented. I first verify that the client doesn't crash, then I verify that the behavior is expected. It's not a fully automatic process, but with some packet captures from a Notchian server it could be.

nickelpro:
Related to the bug, I'm writing a Burger topping to extract the values for all minecraft-data supported versions. That will provide the particles.json we need, then we just need to decide how best to incorporate that into protocol.json

nickelpro:
@rom1504 I have the particle ids for all versions except 1.7.10 (because the "identify" topping doesn't work pre-1.8), would it be more useful to format them as an array? Or as an object that maps the strings to the integer id?

And then the real question is what do you think the best way to incorporate them into the protocol is? Now that I have the ids I can just automate the leg work of correcting the protocol with hard coded integers, which is probably the most compatible option.

rom1504:
How many different kind of particle data is there ?
But yes, it makes sense to try and put all the different kind of particle data in the protocol.
If you have things like id and names it could also be useful as a separate particles.json file.

PR #319, opened 2020-07-29
Add particles.json for all PC versions

nickelpro:
First part of fixing #316

I think the answer for fixing the protocol is to just to fix the particleData type with appropriate hard coded indexes. It's a one time thing and doesn't add any new complexity to protocol.json

If it's not clear from the files themselves, each string's array index is its particle ID. I wasn't sure if this was better as an array, an object, or if I should ship both. So I stuck with the simplest thing.


rom1504:
Is there any data available on particles from burger ?
If not I think it would still be better to do an array of {"id": 0, "name": "the name"} to keep the same style as the rest of the Minecraft data files. And could you also add a simple schema file (you can put it in the schema dir), so it can be used to check the consistency of these values? (for future PRs)

nickelpro:
@rom1504 All of this data is from burger https://github.com/Pokechu22/Burger/pull/14

I'll update the schema to and add a schema file as requested

rom1504:
if you make a script to convert from burger format (only if it's a different format), could you put it in https://github.com/PrismarineJS/burger-extractor ?

nickelpro:
Added schema and updated the particles.json files.

I'm unfamiliar with JS workflows and don't have one set up, but I'll try to figure it out tomorrow and get an extractor to burger-extractor. I'm doing a really trivial transformation on Burger's output though, so anyone can feel free to just push it instead of me.

In python:
```Python
def Burger2MCD(in):
  return [{"id": idx, "name": name} for idx, name in enumerate(in)]
```

rom1504:
can you add particles in https://github.com/PrismarineJS/minecraft-data/blob/master/tools/js/test/test.js#L14 so the test will run ?

PR #324, opened 2020-08-10
Create Loot Data

TheDudeFromCI:
This contains pretty simple parses of the loot tables for all supported Minecraft blocks and entities, via the [loot table plugin.](https://github.com/PrismarineJS/prismarine-loottable)

I wasn't sure what kind of information to pull over, so I grabbed some basic information like requires silk touch, or player kill, as well as the estimated drop chance, before any functions are applied.

Are there any other important elements that should be noted for drops before merging?

### DIFF
--- a/data/pc/1.14.4/loot.json
+++ b/data/pc/1.14.4/loot.json
@@ -1,0 +1,7381 @@
+[
+  {
+    "block": 304,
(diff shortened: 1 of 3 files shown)


rom1504:
not passing the tests, please fix

Issue #323, opened 2020-08-06
Start adding 1.16.2 changes


rom1504:
Seems that it's almost there and there are a few protocol changes https://wiki.vg/index.php?title=Pre-release_protocol&action=history
Could be nice to have 1.16.2-pre3 protocol

nickelpro:
I'm working on this and I'm fairly certain we need a new native type.

1.16.2 encodes block changes as varint based bit fields, where the 12 LSBs are always the chunk section offsets but the MSBs are based on the block id, and thus variable. We need a type that decodes as a varint, shifts the offsets out, and uses the whatever is left as a value.

rom1504:
what packet ?

nickelpro:
https://wiki.vg/Pre-release_protocol#Multi_Block_Change

rom1504:
ok I would encode that as an array of varint and let the user do appropriate bit operations.

nickelpro:
I've done it and intend to PR, but I don't really like that solution. Functions that worked perfectly fine with MultiBlock change before 1.6.2, and conceivably could still work, are now broken. We decode all other bitfields and weirdness (topbitsetarray) at the packet level, why not this one?

rom1504:
topbitsetarray was necessary, here it's not really necessary.
However if you have an idea of a reasonable native type (like not overly specific) we could add, please say.
What could we add except some kind of varint bitfield ?

nickelpro:
In my original version I just had a native type called "varbitfield". It worked identically to a bitfield except it had a "remainder" field that would be left over after the shifts.

It's really no more necessary than topbitsetarray. If you wanted to you could decode topbitsetarray as a rest buffer and leave the decode up to the user. At some level of granularity, all packets are just rest buffers that we special case from. I think maintaining the consistent interface to a given packet is more important than the difficulty of initially implementing the native types.

rom1504:
varbitfield might be ok
The cost in "initially implementing the native types" is important to take into account not for the current implementation of protodefs but for future ones.
If we have few native types, that makes it easier to implement new and better protodef implementation (or in other languages)

in this case it could make sense to introduce it. Could you have a look how it would look like in js?

nickelpro:
I took a good long look at node-protodef and node-minecraft and decided I have no idea what I'm doing. If you understand how bitfields and varints work, then this type is trivial though. Typical generated code for a bitfield right now looks like this:

```
"example": [
  "bitfield",
  [
    {
[...]

PR #326, opened 2020-08-14
Protocol support for 1.16.2

nickelpro:
This adds protocol support and data paths for all unchanged data for 1.16.2, entities and items changed so they're not included.

This uses the original solution of varints proposed in https://github.com/PrismarineJS/minecraft-data/issues/323, while I think a more complete description of the data would be useful, I don't need it personally so I'd rather get this in tree.

This has experienced some limited testing via mcd2cpp, where it parsed common traffic fine.


nickelpro:
Is there something blocking this that I could be working on?

rom1504:
no nothing blocking, just didn't have time to review
trying to look now, but it's hard to diff

CactusKing101:
thx

PR #328, opened 2020-08-20
Missing entities for 1.16.2

DrakoTrogdor:


nickelpro:
Piglin Brutes got added in 1.16.2, so just adding 1.16.1 to datapaths will result in incomplete data. Better to copy the entities.json to 1.16.2 and set datapaths to that. At least that way it's setup to be updated.

rom1504:
It would be better to extract data the right way. I started a guide about
that in another pr

DrakoTrogdor:
Ya, after I added the forked minecraft-data to my project and started to really try things I noticed the ID's were all borked, I'll close this and try to figure out extractors as per rom1504

PR #333, opened 2020-09-10
Add support for 1.16.3

nickelpro:
1.16.3 is a [bug fix release](https://www.minecraft.net/en-us/article/minecraft-java-edition-1-16-3), there are no protocol or data changes except for version bumps. Confirmed with [BurgerVitrine](https://pokechu22.github.io/BurgerWebViewer/).

This patch adds the necessary version.json and back references everything else in dataPaths.json


rom1504 (changes requested):
can you please add it there too https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/common/versions.json ? that's what triggers tests

nickelpro:
Force pushes are for forgetting to actually add the new version to dataPath.json and forgetting to add the supported version to the ReadMe. Nothing to see here, just a single programmer struggling with basic stuff.

nickelpro:
Added to versions.json, tests pass

rom1504 (approved):

PR #334, opened 2020-09-12
Added a command entry for almost every mc-version > 1.13

Miro-Andrin:
I was working on command sport for [Feather-rs](https://github.com/feather-rs/feather), a rust based Minecraft server. I wrote some code that took data from [minecrafts data generators](https://wiki.vg/Data_Generators) and generated rust code from it. I was thinking about extending the data so that i could auto generate testcases, and maybe even give grammars for the sub parsers. I believe other projects could benefit from the data, so it makes sense to share it here.

I have currently only added testcases for some of the parsers. I am planning on adding more data as feather develops. 

The data was generated using:  https://github.com/Miro-Andrin/mc-data-command-generator.git


rom1504:
could you please add a line in the data and extractor tables?

rom1504:
Ah and I forgot one last thing: https://github.com/PrismarineJS/minecraft-data/blob/master/tools/js/test/test.js#L14 can you add commands there so the test run ?

rom1504:
Thanks!

PR #336, opened 2020-10-03
Add biomeLogin for 1.16.2 & 1.16.3

GroobleDierne:
The data are the same for both versions


IdanHo:
i suggest you add 1.16/1.16.1's as well, and update flying-squid to use that instead of the hardcoded one it has right now

GroobleDierne:
1.16 and 1.16.1 do not send the part worldgen/biome

rom1504:
Is enough to fill out the default values in the login packet ?

IdanHo:
It does: https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/login.js#L239

GroobleDierne:
> Is enough to fill out the default values in the login packet ?

No, the example will look like the following
```javascript
const mc = require('minecraft-protocol')
const biomeLogin = require('minecraft-data')('1.16.3').biomeLogin

const server = mc.createServer({
  'online-mode': false, // optional
  encryption: false, // optional
  host: 'localhost', // optional
  port: 25565, // optional
  version: '1.16.3',
  serverBrand: 'test'
})

server.on('login', function (client) {
  
  const w = {
    piglin_safe: {
      type: 'byte',
      value: 0
    },
    natural: {
      type: 'byte',
      value: 1
    },
    ambient_light: {
      type: 'float',
      value: 0
    },
    infiniburn: {
      type: 'string',
      value: 'minecraft:infiniburn_overworld'
    },
    respawn_anchor_works: {
      type: 'byte',
      value: 0
    },
    has_skylight: {
      type: 'byte',
      value: 1
    },
    bed_works: {
      type: 'byte',
      value: 1
    },
    has_raids: {
      type: 'byte',
      value: 1
    },
    effect: {
      type: 'string',
      value: 'minecraft:overworld'
    },
    logical_height: {
      type: 'int',
      value: 256
    },
    ultrawarm: {
      type: 'byte',
      value: 0
    },
    has_ceiling: {
      type: 'byte',
      value: 0
    },
    coordinate_scale: {
      "type": "double",
      "value": 1
    }
  }
  client.write('login', {
    entityId: client.id,
    gameMode: 0,
    previousGameMode: 255,
    worldNames: ['minecraft:overworld'],
    dimensionCodec: {name: '', type:'compound', value: {
      'minecraft:dimension_type': {
        "type": "compound",
        "value": {
          "type": {
            "type": "string",
            "value": "minecraft:dimension_type"
          },
          "value": {
            "type": "list",
            "value": {
              "type": "compound",
              "value": [
                {
                  "name": {
                    "type": "string",
                    "value": "minecraft:overworld"
                  },
                  "id": {
                    "type": "int",
                    "value": 0
                  },
                  "element": {
                    "type": "compound",
                    "value": w
                  }
                },
                {
                  "name": {
                    "type": "string",
                    "value": "minecraft:overworld_caves"
                  },
                  "id": {
                    "type": "int",
                    "value": 1
                  },
                  "element": {
                    "type": "compound",
                    "value": w
                  }
                }
              ]
            }
          }
        }
      },
      'minecraft:worldgen/biome': biomeLogin,
      dimension: {type: 'list', value: {type: 'compound', value: [w]}}
    }},
    dimension: {type: "compound", name: "", value: {w}},
    worldName: 'minecraft:overworld',
    viewDistance: 10,
    difficulty: 2,
    hashedSeed: [0, 0],
    maxPlayers: server.maxPlayers,
    reducedDebugInfo: false,
    enableRespawnScreen: true,
    isHardcore: false,
    isFlat: false
  });
  
  client.write('position', {
    x: 0,
    y: 1.62,
    z: 0,
    yaw: 0,
    pitch: 0,
    flags: 0x00
  });
})

```

rom1504:
are you sure, can you post a dump of the whole login packet ? I think there is more
but anyway, I propose to store here a whole example of the login packet, that'll simplify that code much more

rom1504:
@IdanHo the code you posted is for 1.16.1, they added much more in 1.16.2

GroobleDierne:
@rom1504 generated by Paper 1.16.3 https://pastebin.com/DisB3TQG

rom1504:
can you generate it using vanilla ? we want the default behavior to be like vanilla

rom1504:
as you can see from this dump, it's not only the biome part that is big but at least the whole dimension codec, and even the dimension field

GroobleDierne:
I used vanilla for 1.16.2 and it was the same result

rom1504:
ok so I think it makes sense to dump a whole example of the whole packet here, and extract the good parts in examples and tests in node-minecraft-protocol

rom1504:
that means it works for 1.16 and 1.16.1 too

GroobleDierne:
I'm not sure I understand, put the whole login packet in MCData or only the dimensionCodec and dimension fields ?

rom1504:
the whole login packet

GroobleDierne:
Ok

rom1504:
could you add it for 1.16 and 1.16.1 too ?

GroobleDierne:
Yes I'm testing for 1.16

rom1504 (approved):

PR #344, opened 2020-11-13
Fix 1.16.2 block states

DeltaEvo:
The enum values are uppercase in the 1.16.2 blocks file, there is also a new "direction" type that is just an enum, and some values dosen't match mojang name

For example the same block in 1.16.2 and 1.16.1 https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.2/blocks.json#L1810-L1818
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.16.1/blocks.json#L1956-L1965

### DIFF
--- a/data/pc/1.16.2/blocks.json
+++ b/data/pc/1.16.2/blocks.json
@@ -1807,7 +1807,7 @@
     "states": [
       {
         "name": "facing",
-        "type": "direction",
+        "type": "enum",
         "values": [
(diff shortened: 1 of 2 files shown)


DeltaEvo:
I removed the draft status since I want the CI to run on this PR

rom1504 (approved):

Issue #330, opened 2020-08-21
Missing Netherite recipes?


njt1982:
There doesn't seem to be any recipes for the Netherite tools:

https://github.com/PrismarineJS/minecraft-data/blob/526f69d2993ce3f4b51726ec29fae6f0036e6ce9/data/pc/1.16.1/items.json#L3639-L3661

Is this something that needs prodding in the burger extractor?

Issue #290, opened 2020-05-31
Smelting/Furnace data


njt1982:
Hi,

Maybe I'm missing something here...

Is there any data about how things like Iron Ore gets turns into Iron Ingots? It's obviously not a crafting recipe...

njt1982:
Have raised this in Burger too as it seems to be missing "upstream"...
https://github.com/mcdevs/Burger/issues/35

Seems to be intentionally excluded...
https://github.com/mcdevs/Burger/blob/fdff962aeb1aa0351fc222e005af3fa9248345fb/burger/toppings/recipes.py#L136

rom1504:
Yeah basically we would need to check if these recipes could fit in the same format.
If they do we could add them there.
We would probably need to tag them though

njt1982:
I'd be inclined to put them in a smelting.json (or similar). They have different structures; in/out/timing. There is no shape, no "count":

```
{
  "type": "minecraft:smelting",
  "ingredient": {
    "item": "minecraft:iron_ore"
  },
  "result": "minecraft:iron_ingot",
  "experience": 0.7,
  "cookingtime": 200
}
```

Thats the raw data from the iron_ingot.json in the minecraft server.jar

rom1504:
Sounds good to me.
If you're interested by this, feel free to propose changes in the pipeline so we can have that :
* In burger to have a recipe type option
* In burger extractor to put the recipe in the right format
* In minecraft data to define a good schema and put the data (at least for the last version)
* In node-minecraft-data to expose that data
* In prismarine-recipe to use that data
* In mineflayer to actually implement these recipes

Before doing all this though, could you please check how many recipes that is ? If it's just 2 or 3, we could maybe do something different

njt1982:
Hi,

Based on https://github.com/mcdevs/Burger/issues/35 it's ~50 recipes for smelting. But @Pokechu22  raised an interesting point about smokers and blast furnaces too. They also made this suggestion:

> It might be easier for minecraft-data to use the data generator system (or directly extracting the JSON files from the minecart jar) themselves, and bundle it into their own recipes.json or equivalent, rather than that data first being put into burger and then put into minecraft-data.

rom1504:
I'm fine with that. Won't change too much the process I described above.

Feel free to work on it then. Probably worth it to do all types of recipes and not only smelting

extremeheat:
[...]

PR #345, opened 2020-11-14
Revert "Adding Minecraft Item Browser project"

njt1982:
Reverts PrismarineJS/minecraft-data#329

Hi - quick PR to remove the link... I no longer use Prismarine for a data source; I'm parsing the JAR file manually.

The combination of #290 and #330 meant that I was missing a lot of useful recipe data. I honestly tried to find a way to get the data into the toolchain for this dataset, but I couldn't figure out what I needed to change and where.... (combined with the Burger project resisting adding more recipes).

Thanks for all the help with the data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,6 @@ These are the projects that use minecraft-data:
 | [McEx](https://github.com/hansihe/McEx) | Elixir | server | blocks |
 | [VoxelGamesLib](https://github.com/MiniDigger/VoxelGamesLib) | Java | minigames server library | blocks, items |
 | [Phpcraft](https://github.com/Phpcraft/core) | PHP | all-around library | blocks, items, entities, protocol |
-| [Minecraft Item Browser](https://njt1982.github.io/minecraft-item-browser/) ([Repo](https://github.com/njt1982/minecraft-item-browser)) | Javascript | Web-based recipe search tool | blocks, items, recipes |
 
 ## Extraction
 


rom1504:
What do you use as format after extracting from the jar ?

njt1982:
It parses data in the JAR into JSON in a very similar structure to Prismarine...

Script:
https://github.com/njt1982/minecraft-item-browser/blob/master/scripts/parse_jar.js

Output (items / recipes / titles):
https://github.com/njt1982/minecraft-item-browser/tree/master/public/js

I essentially had the site working under Prismarine as I needed but it was simply missing data around recipes for things like Smelting... Originally I used this [update.js script](https://github.com/njt1982/minecraft-item-browser/blob/master/scripts/update.js) which is super simple; it basically copied the data out and massaged it slightly for my needs.

Items get parsed into structures like this:
```
{
  "id":0,
  "name":"acacia_planks",
  "displayName":"Acacia Planks",
  "texture":"iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAA1ElEQVQ4y4VTIRICQQzLcxgwDAaJxCBQOBSaf5xD4/gAH0DysWWyM9kJoYDodbftbJs0h8dp3e7HVaN/njcf9it+3S8aeNAjNAZvh+V4NPO6s4aGqlBeRbJ8pE/Az2U378az7tN21k05nT1OD2Fkh8SZMYfCHO9vEJSQdz4Squrh+LyDCK2IdF6QWJODzGcNfM/e5Z8WtEr46nLcXKF7cYSK3Wr3mR8TOMZKB55zTcjGBMmFOjhuceO1+KbxHLdaZ9eBF+Z/4Dqo+BkcJDbnwzmp9PECHCoz2n4Vz4IAAAAASUVORK5CYII="
}
```

Recipes now define their "type", eg:

```
{"type":"crafting_shaped","ingredients":[0],"inShape":[[0,null,0],[0,0,0]],"id":0,"result":{"id":1,"count":1}},{"type":"crafting_shapeless","ingredients":[0],"id":1,"result":{"id":2,"count":1}}
```

There is still a bit of work to do... Eg Tagged items such as "planks" (which can be any type of plank)... currently pulls just the first plank type. That's "ok", but I'd like to extend that definition a bit.

If there is anything in there that can help this project, feel free to "be inspired" by it 😉

rom1504:
Why not import your recipe file inside mcdata ?
It's not necessary for us for data to come from burger, we can have other
kind of extractors

njt1982:
@rom1504 would be happy to help on that if I had the foggiest idea how to 😉

I've searched this repo for McData and it seems to be mentioned on the homepage as a link to a repo + several issues mention it. The McData repo says `TODO: Add description` and has no Wiki.

Without documentation for those unfamiliar with the guts of this project its an almost cliff-like learning curve. To be honest, this is the reason I stopped looking into this RE issue #290 in the first place... I saw this project used burger for recipes, I went to look at fixing up burger to produce recipes for smelting and was blocked on that by the project maintainer... There wasn't clear direction/steps for me to take to open a PR to help out.

I do genuinely want to help, so please dont take this critisism as any kind of attack... If you can point me in the right direction I'll happily take a look.

rom1504:
Indeed what I was suggesting was simply to put the file you already generated in minecraft data and that's it. The generator is your own code.
The only thing that would be good is using the same format as is currently used for crafting recipes.
For other kind of recipes a similar format would make sense (probably in another file)

The benefit of doing that is increasing the change this data will stay maintained and updated for new versions, and more people can benefit from it.

I agree the extractors are not as well documented as they could be, they're not really organised yet in a single set of extractors. So for now we use anything that works, including the jar extractor code you wrote for example.

Does that make sense to you ?

njt1982:
Ohhh!! I was under the impression there was some kind of automated/CI workflow for getting data consistently in via "known" scripts...

In that case I could probably knock something custom up to pull this data directly from a JAR using a NodeJS script...

The one I use at the moment modifies the structure slightly from this repo... I'll see if I can alter the script to keep it consistent. Did you want all the recipes in recipes.json or split out into "types"? (smelting, etc).

rom1504 (approved):

PR #348, opened 2020-12-27
Added map icons

FalcoG:
This adds map icons for version 18w10a and newer.

Sources:
https://minecraft.gamepedia.com/File:Map_icons.png (last updated at 18w10a)
https://minecraft.gamepedia.com/Map#Map_icons
https://wiki.vg/Protocol#Map_Data


TheDudeFromCI (approved):

rom1504 (changes requested):
Hi,
This looks like a good new data to have.
Can you please add a description of what it is in the readme?

I think it's not great to support only one snapshot for this data.
Did it change a lot in the past?

Maybe the wiki history can provide some information if this changes often

can you also add there https://github.com/PrismarineJS/minecraft-data/blob/master/doc/add-data-new-version.md how you got this data

FalcoG:
@rom1504 I've now added the previous versions of the map icons. It only changed a few times. 

A description of where to look up the map icons for a Minecraft version has also been added.

rom1504:
Is this data outdated for recent versions ?

rom1504:
(if not, might as well put it in 1.16.4 folder and remove this snapshot folder)

FalcoG:
Reason I've put it in 18w10a (released March 6, 2018) is because that was the last time the icons have changed. Would it be better to duplicate the file for the latest version?

rom1504:
We don't really do "keep only the latest update" in this repo.
Ideally every folder would contain the data for the version, even if there is no diff.

I don't really like adding a new version with all data files missing except one. There's no added benefit since we can't use this version, too much data is missing.

That's why I was suggesting to add it only in 1.16.4 and reference it from other versions. Or add it in the nearest major. Or add it for every version.

rom1504:
And are you sure minecraft wiki is correct regarding the last update of this data ?

FalcoG:
> And are you sure minecraft wiki is correct regarding the last update of this data ?

Yeah by using minecraft-jar-extractor it reveals that map_icons.png from 1.16 is exactly the same as 18w10a, 1.13, 1.14 etc.

The sprite changes as I go further back; for example in 18w09a it's different.

I've compiled a small changelog:
1.13 (18w10a): banners and red cross added
1.11: tiny white dot, woodland mansion and temple monument added

> I don't really like adding a new version with all data files missing except one. There's no added benefit since we can't use this version, too much data is missing.

Alright, I'll remove the snapshot version and move the file to 1.13 because that's the latest major where this was added.
For 1.16.x I'll make a copy of the data to the 1.16 folder and have the references there point to 1.16 only.

rom1504 (approved):

rom1504:
Great, thanks!

Out of curiosity, how do you want to use this data ?

FalcoG:
I intend to add support for maps to flying-squid and this change should make calling the map icons easier.

So far I've managed to process the map data file, add map icons at specific locations and have a dynamic player cursor displayed.

The player cursor/icon responds to player movement and it also supports rotation. When the player goes outside of the map area, the cursor will have to change to another icon.

If you're interested you can take a peek at the code here: https://github.com/FalcoG/flying-squid/commit/8b2b40b0f9f37662bbff314dd163212cefd791fd

rom1504:
cool.
You will need to load the data in https://github.com/PrismarineJS/node-minecraft-data too so flying-squid can access it

rom1504:
actually just did it https://github.com/PrismarineJS/node-minecraft-data/commit/5a32c8661b927d706aa08a1ce5684ab6c3a97564

you can just update mc data to 2.71.0 and you can access that data

FalcoG:
Awesome, thanks for updating!

PR #349, opened 2021-01-10
Fix recipes

### LINKED ISSUE #231: Pickaxe recipe upside down?
https://github.com/PrismarineJS/minecraft-data/blob/a3bdea5217e3c027441207f8fcdea7b011c132bf/data/pc/1.13.2/recipes.json#L6745-L6769

According to https://minecraft.gamepedia.com/Pickaxe the iron should be along the top, not along the bottom.

Why is the recipe vertically inverted?

Karang:
* Fix #231
* Add audit script
* Fix 1.9 and 1.10 recipes by copying 1.11


Karang:
In addition to this, more meaningful tests should be added to mineflayer. For instance the crafting test could craft a pickaxe, that would have caught this error earlier.

rom1504:
yeah we could take https://github.com/PrismarineJS/mineflayer/blob/229ae9443e0f849cfbdb3dd1a896d475e6afd0b7/test/externalTests/crafting.js and make one with more item to craft (or everything like https://github.com/PrismarineJS/mineflayer/blob/229ae9443e0f849cfbdb3dd1a896d475e6afd0b7/test/externalTests/digEverything.js )

rom1504:
> Fix 1.9 and 1.10 recipes by copying 1.11

are you sure there's no additional things in 1.11 ?

rom1504:
we probably also need to fix the extractor but that's independent

Karang:
> > Fix 1.9 and 1.10 recipes by copying 1.11
> 
> are you sure there's no additional things in 1.11 ?

No i'm not, but its better to have more than having missing recipes

rom1504:
what is missing in 1.9 and 1.10 exactly?

Karang:
> what is missing in 1.9 and 1.10 exactly?

What i found is it was missing the main pickaxes recipes https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.10/recipes.json#L5431
(there is only the one that combine 2 used pickaxes)

But looking at the diff, there is actually a lot more recipes in those files than in other versions (27000 lines in 1.10 vs 7000 lines in 1.11 and 1.12). So they are both missing something... The only way to get it right would be to re-extract all recipes for versions below 1.13. If you want I can revert the commit for 1.9 and 1.10 in order to not delay the fixing of the upside down recipes.

rom1504:
yeah I think it's better to keep the 2 things separate.
for other versions than 1.9 and 1.10 you should reversed the order right ?
let's just fix this order problem in this PR

for the missing/additional recipe problem, indeed the right fix would be to rerun burger on those
we could just use the 1.11 recipe file if that's difficult, but in that case I'd just delete the 1.9 and 1.10 files to avoid further confusion

rom1504:
and I guess just exclude these 2 versions in your new test for now

rom1504 (approved):

Issue #325, opened 2020-08-10
Missing new 1.16.0 enchantment


IdanHo:
No new enchantments were added since 1.13.2 (and as such no version since 1.13.2 had an enchantments.json file), and as such were missing the new Soul Speed enchant added in 1.16.0

nickelpro:
No version since 1.13.2 has an enchantments field included in dataPaths.json at all. It's more correct to say no version since 1.13.2 has ported the enchantment data forward at all, largely because it's all manual work and enchantment data is little used. This was attempted to be addressed by https://github.com/PrismarineJS/minecraft-data/pull/287 and the associated PR on the wiki extractor, but the wiki doesn't provide all needed data for the MCD schema.

The immediate issue of hand porting enchantments.json forward for 1.14+ is easy and I assume Rom will accept a PR if you do that. The more complete answer is to write a new topping for Burger that extracts this data.

nickelpro:
I've identified the class for extracting the data and intend to take a wack at building the burger extractor sometime this week when I have a minute. As long as we're doing this we can grab some other stuff MCD doesn't track like rarity and equipment slot.

@rom1504 Is it ok to add schema fields as long as we maintain old ones?

IdanHo:
I'd be happy to open a pull request for the enchantments myself, but i can wait as well for the extractor if youd like.

nickelpro:
If it's not urgent, give me two days and I should have something working. It's a little harder than the two-line extractor you'd want it to be because MC changed the way these pseudo-enumerated types worked around 1.13, so getting support for all the versions Burger works on can be a little tricky.

rom1504:
Sure it's fine to add new fields

nickelpro:
So I've run into a little bit of a snag

The sequential Window ID enchantment IDs are not the same as the deprecated NBT IDs for enchantments. For example in 1.13.2 the NBT ID for "punch" is 49 (https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.13.2/enchantments.json#L118), but the Window ID is 23. The NBT IDs are completely deprecated and any assignement in later versions will be arbitrary, and in earlier versions we need to add a "window_id" field. The Soul Speed enchantment is a good example, it has no NBT ID field listed anywhere in the jar since it came about after they were removed.

What's the best way to indicate a deprecated field in MCD? Leave it blank? Set to zero? Remove it entirely?

rom1504:
[...]

PR #353, opened 2021-01-16
Fix 1.13 enchants and add new 1.16.4 enchants

Karang:
The enchants of 1.13 had wrong ids and some bad names according to https://wiki.vg/index.php?title=Protocol&oldid=14889

New enchants were only added in the 1.16.4 version according to https://wiki.vg/index.php?title=Protocol&oldid=16091 and https://wiki.vg/index.php?title=Protocol&oldid=16317

This PR fix both issues

### DIFF
--- a/data/dataPaths.json
+++ b/data/dataPaths.json
@@ -837,7 +837,7 @@
       "biomes": "pc/1.16.2",
       "effects": "pc/1.16.1",
       "items": "pc/1.16.2",
-      "enchantments": "pc/1.13.2",
+      "enchantments": "pc/1.16.4",
       "recipes": "pc/1.16.2",
       "instruments": "pc/1.16.1",
       "materials": "pc/1.16.2",
--- a/data/pc/1.13/enchantments.json
+++ b/data/pc/1.13/enchantments.json
@@ -17,6 +17,7 @@
       "fire_protection",
       "projectile_protection"
     ],
+    "category": "armor",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
@@ -40,6 +41,7 @@
       "protection",
       "projectile_protection"
     ],
+    "category": "armor",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
@@ -59,6 +61,7 @@
       "b": 5
     },
     "exclude": [],
+    "category": "armor_feet",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
@@ -82,6 +85,7 @@
       "protection",
       "projectile_protection"
     ],
+    "category": "armor",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
@@ -105,6 +109,7 @@
       "blast_protection",
       "fire_protection"
     ],
+    "category": "armor",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
@@ -124,6 +129,7 @@
       "b": 30
     },
     "exclude": [],
+    "category": "armor_head",
     "treasureOnly": false,
     "curse": false,
     "tradeable": true,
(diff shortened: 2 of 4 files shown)


nickelpro:
Those ids aren't wrong, they're the NBT ids from 1.13, which were deprecated in subsequent versions. The window IDs are seperate, this was discussed briefly in https://github.com/PrismarineJS/minecraft-data/issues/325.

I think the correct answer requires a schema change, because the NBT ids are important for dealing with world data in 1.13 and the Window IDs are obviously important for correctly interpreting the protocol.

Karang:
> No, the NBT IDs are not used at all in later versions. They only exist in "Fix" classes that map old enchantment data to the new format, for updating existing world files. Soul Speed did not exist in old versions and therefore doesn't even have an ID listed there.

This is what you commented on the issue. If the nbt id is only used for conversions, it should be in a separate data, like what we've done for https://github.com/PrismarineJS/prismarine-schematic/blob/master/lib/legacy.json

Do you know of any place where the id is used and expected to be a nbt id ? We need this changes anyway for the protocol.

nickelpro:
You need both to identify data in <= 1.13.2, because like I said, it's the NBT ID. IE, all slots with an enchanted item, the [optionalNBT field](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.13.2/protocol.json#L32-L70) will identify that enchantment with the NBT id. After 1.13.2 the string ID is used.

Karang:
> You need both to identify enchantment data in <= 1.13.2, because like I said, it's the NBT ID. IE, all slots with an enchanted item, the [optionalNBT field](https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.13.2/protocol.json#L32-L70) will identify that enchantment with the NBT id. After 1.13.2 the string ID is used.

I just checked, on 1.13.2 nbt enchant id is a string:
```
// on a book:
client<-server: play.set_slot :{"windowId":0,"slot":38,"item":{"present":true,"itemId":714,"itemCount":1,"nbtData":{"type":"compound","name":"","value":{"StoredEnchantments":{"type":"list","value":{"type":"compound","value":[{"lvl":{"type":"short","value":1},"id":{"type":"string","value":"minecraft:mending"}}]}}}}}}
// on an item:
client<-server: play.set_slot :{"windowId":1,"slot":2,"item":{"present":true,"itemId":786,"itemCount":1,"nbtData":{"type":"compound","name":"","value":{"RepairCost":{"type":"int","value":1},"Enchantments":{"type":"list","value":{"type":"compound","value":[{"lvl":{"type":"short","value":1},"id":{"type":"string","value":"minecraft:mending"}}]}}}}}}
```

nickelpro:
Derp, I'm getting versions mixed up, 1.12.2 was the last version that used enchant IDs. 1.13 was post-flattening and used the string IDs. This should be fine for 1.13 then, but older versions will still be wrong.

Karang:
> Derp, I'm getting versions mixed up, 1.12.2 was the last version that used enchant IDs. 1.13 was post-flattening and used the string IDs. This should be fine for 1.13 then, but older versions will still be wrong.

So its fine, this PR doesn't change data < 1.13

nickelpro:
Yep, ship it. I think it's weird that the meaning of the ID field changes between versions now though. Since Window IDs exist in older versions too, but the ID field now only identifies Window IDs in 1.13.2+

rom1504:
Maybe we can have 2 fields ?
Where do we use this in PrismarineJS modules today ?

Karang:
In 1.12.2, there is no enchant id sent in the window (because the enchantments are hidden it always send -1).

AFAIK the only place where we use enchantments that have a different windowid than their nbt id is in p physics https://github.com/PrismarineJS/prismarine-physics/blob/master/index.js#L634 and we already account for the change from numeric to string.

rom1504 (approved):

PR #357, opened 2021-01-18
Add missing materials

mat-1:
Netherite tools weren't in materials.json, and a lot of 1.14 blocks had unknown_material. I fixed all of these except bamboo because bamboo is weird


Karang (approved):
This looks good @rom1504

rom1504 (approved):

rom1504:
Looks good but I think we should have a good audit material making sure the items referenced in materials are really tools.

PR #358, opened 2021-01-21
More data for biomes

Karang:


rom1504:
Missing files in versions really means "missing" as in "we did not run the extraction for this version".
I'd better keep that semantic true and keep the identical versions of files.
I'm aware that means a larger repo and larger assets but I think the simplicity it brings is worth it.
The data path file is meant to compensate for missing files and not to try to gain some space.

Note that we could consider storing only the last version in the npm package and generated diffs for other versions (and then a tool to regenerate the full file at read time) if we care about decreasing the size of node_modules.

data/dataPaths.json
```diff
@@ -617,7 +617,7 @@
       "enchantments": "pc/1.13.2",
       "version": "pc/1.15",
       "blocks": "pc/1.15.2",
-      "biomes": "pc/1.15.2",
+      "biomes": "pc/1.15",
```
rom1504, line 620: This seems to be incorrect isn't it ?
It seems you updated that file too

rom1504 (changes requested):

data/dataPaths.json
Karang: I've put a biome file in every major version, so 1.15 have its own biomes.json. How is it incorrect?

rom1504: Ah right, I misread this, I thought this was for 1.15.2

rom1504 (approved):

PR #361, opened 2021-02-17
Fix 1.16 blockCollisionShapes

Karang:
I reran the extractor for 1.16 shapes (thanks @u9g for fixing the mod and making it easy to run it)

I formatted it manually to minimize the diff so we can understand the changes. In the future, it would be nice to reformat all blockCollisionShapes.json with a script...


u9g:
I can pr something to the script to add postprocessing?

rom1504 (approved):

PR #362, opened 2021-02-17
Add tag.yml action

Karang:


rom1504:
https://github.com/christophebedard/tag-version-commit#inputs default regex will not work

rom1504 (approved):

PR #363, opened 2021-02-18
Fix incorrect 1.7 definitions

SiebeDW:
The 1.7 protocol definition is using incorrect fields according to [wiki.vg](https://wiki.vg/index.php?title=Protocol&oldid=6003#Teams)
This pull request will use the correct fields.

### DIFF
--- a/data/pc/1.7/protocol.json
+++ b/data/pc/1.7/protocol.json
@@ -2040,7 +2040,16 @@
             },
             {
               "name": "scoreName",
-              "type": "string"
+              "type": [
+                "switch",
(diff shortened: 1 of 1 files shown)


rom1504:
How did you test this ?

SiebeDW:
I used a node minecraft protocol proxy to connect to a 1.7.10 mc server. I could not connect to it because it was parsing the scoreboard packets wrong. When I used the correct definition I was able to connect without any errors.

rom1504 (approved):
ok, thanks

PR #366, opened 2021-02-22
Add ItemMultiplier, BookMultiplier, Exclude (to ones missing them)

u9g:


rom1504 (changes requested):
Can you put these js files in an extractor repo instead ?

u9g:
> Can you put these js files in an extractor repo instead ?

done https://github.com/u9g/enchant-multiplier-injector/tree/main

Karang (approved):

rom1504:
Same comment there, add a link to that extractor to https://github.com/PrismarineJS/minecraft-data/blob/master/doc/add-data-new-version.md (and also to the table in the readme)

PR #371, opened 2021-02-23
Propagate catgeory, maxLevel

u9g:


Karang (approved):

rom1504:
@u9g that looks fine but could you add some notes on how to generate that in guide file in doc ?
It's important that we know how to generate that in the future if you're not available

rom1504:
https://github.com/PrismarineJS/minecraft-data/blob/master/doc/add-data-new-version.md

u9g:
> @u9g that looks fine but could you add some notes on how to generate that in guide file in doc ?
> 
> It's important that we know how to generate that in the future if you're not available

Its not needed, any newer versions that come out, the extractor natively comes with these, this was just to bring the properties to older enchantment.json s

rom1504:
"the extractor" is not documented in this .md file so that's what need to be done

PR #377, opened 2021-02-25
add treasureOnly, weight, curse, tradeable, discoverable, rem multiplier

u9g:


u9g:
this doesn't need any info because the newest enchantments.json comes with all this data

Karang:
@u9g it seems like you added an extra out.json. Otherwise, LGTM.

rom1504 (changes requested):
Can you adapt the schema ?
Is this already explained in the extraction guide ?

u9g:
> @u9g it seems like you added an extra out.json. Otherwise, LGTM.

done

rom1504 (approved):

Issue #368, opened 2021-02-22
Block variations


CyberPatrick:
https://github.com/PrismarineJS/minecraft-data/blob/381f51e9c7481098c5b0690fd074b4d943d7171d/data/pc/1.12/blocks.json#L413-L416

May be instead of adding "(no decay)" in block displayName, create new field with description? Like this:
{ 
   "metadata": 4, 
   "displayName": "Oak Leaves",
   "description": "No decay" 
 },

CyberPatrick:
I started working on that

nickelpro:
If this is the fix we're going with, this needs to be backported to other <1.13 versions

PR #379, opened 2021-02-26
Added new field 'description' and blocks metadata

CyberPatrick:
#368

### DIFF
--- a/data/pc/1.12/blocks.json
+++ b/data/pc/1.12/blocks.json
@@ -375,6 +375,68 @@
     "diggable": true,
     "boundingBox": "block",
     "material": "wood",
+    "variations": [
+        {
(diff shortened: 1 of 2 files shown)


rom1504:
Where does this data come from ? How can it be used ?

CyberPatrick:
From https://minecraft-ru.gamepedia.com/ and from game, i checked every block. For example of logs, it can help user determine difference between log with 0 and 4 metadata. Due to this change, i can count blocks in area by their displayName.

rom1504 (approved):

u9g:
shouldn't this be put into all different blocks.json files?

rom1504:
All files before 1.13 maybe yes

Issue #283, opened 2020-05-05
Write guide on how to provide data for new versions


rom1504:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%242%20Funded-%2300A156.svg" />](https://issuehunt.io/r/PrismarineJS/minecraft-data/issues/283)

using the best extractors we have

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $2.00)

- [<img src='https://avatars1.githubusercontent.com/u/2346494?v=4' alt='rom1504' width=24 height=24> rom1504](https://issuehunt.io/u/rom1504) ($2.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/PrismarineJS/minecraft-data/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

rom1504:
automatize as much as possible

uncovery:
I am working on a bash script that would create the whole process from scratch as much as possible. Please help to expand this: https://github.com/uncovery/minecraft_datascript

rom1504:
These are all the ways that can be used to extract data to provide minecraft data currently : https://github.com/PrismarineJS/minecraft-data/blob/master/README.md#extraction

What I mean with this issues is writing a guide that will explain what are these extractors, their benefit, what we advise to use at the current time.
It would be perfect if we could automatize everything but it's pretty much impossible due to constant changes that Mojang make.

To answer your question:
* Language is generated using minecraft jar extractor
* version file is simple and written manually
* protocol file is copied from previous version then adapted manually using wiki.vg information

There is basically no way to automatize protocol.json generation (I tried)

However for everything else, there are various generators left and right. It is possible to automatize it or at least make it clear how to assemble everything.

uncovery:
Thanks, I get it now. Maybe there is a dual-purpose to my script so people not only use it, but also understand how it works through proper documentation and therefore also understand what the individual extractors' benefit is. It could then be updated to the standard that is currently advised.

I will try to continue on the script to have it do at least what I personally need from it. If anybody wants to add on to it, please be my guest.

uncovery:
You wrote: 

> protocol file is copied from previous version then adapted manually using wiki.vg information

[...]

PR #384, opened 2021-03-21
Adding blast resistance

### LINKED ISSUE #383: Add BlastResistence to Blocks
Would be good to have blast resistence data for each block

Some data to extract: https://minecraft-el.gamepedia.com/Explosion

Moondarker:
Contains current minecraft-data data merged with burger-extracted blast resistance

Closes #383


Moondarker:
Gotta note, blast resistance = undefined (values added as 0 in latest commit) correspond very well with hardness = 0 in the same block, and there are _a lot_ of such blocks in 1.16

Example: Dark Oak Stairs
1.15.2: "hardness": 2, "resistance": 3
1.16.2: "hardness": 0, "resistance": 0

What I mean is, something has got to be wrong with the way 1.16 data is extracted

rom1504:
A bug in burger ?

rom1504:
Did you write some kind of merge script to only add properties and not lose the ones you don't extract?

Moondarker:
Yep, automated it with a node script to merge existing minecraft-data with resistance gathered using burger and a powershell script to make it for every major version
Also, this comment https://github.com/PrismarineJS/burger-extractor/pull/13#issuecomment-803546142 seems to be only partially true, minecraft-data for all versions >=1.13 was clearly extracted using burger-extractor, all versions <=1.12 on the other hand I'm not sure about

Moondarker:
Burger output is affected by https://github.com/Pokechu22/Burger/issues/23, will see if I can make a quick workaround for now.

Moondarker:
This is ready for review, I don't think it can get any better for now. Output of Burger was patched using a list of special cases gathered from 1.16 sources
@rom1504, @Karang, the way minecraft-data was compiled desperately needs to be documented, it will become harder to add new data over time if we can't consistently reproduce results. Also, wiki is a nice source, but by no means a reliable one, especially for any version older than the latest one

rom1504:
Ok will check the pr

About doc I completely agree and the easiest way to do this is that each new pr improve the doc a bit.

So can you add to the "add new version" guide in doc folder what you did here to get blast resistance so the next person can redo the same ?

Moondarker:
Yeah, I just need to clean up a little first. Amount of stuff I collected to get correct values is quite big, but it can be compressed substantially

Moondarker:
Actually, just running burger-extractor with https://github.com/PrismarineJS/burger-extractor/pull/13 should do
However, no matter what I do, I cannot reproduce `blocks.json` from `minecraft-data` by running burger-extractor, some step has to already be missing from "add new version" guide.
I can ofc make a mini-guide on how to merge new data with minecraft-data without re-generating files from scratch, but it's all about egg and chicken paradox, and in future it will lead to problems in determining where different pieces of data came from (if everyone will be merging data like that)

Moondarker:
If we, as a species, lose in some aspects to computers - aspect №1 is data processing

What I mean is - documentation is good, but we can:
1. Automate data merging process from different extractors
2. Improve data extraction from game files instead of getting some parts from wiki

Doing at least one thing of listed above will greatly improve data consistency

rom1504:
yes what you are saying has been tried by several people
https://github.com/PrismarineJS/minecraft-data/issues/283 has some information on this
the last tries on this are:
* https://github.com/extremeheat/minecraft-data-extractor
* https://github.com/PrismarineJS/McDataExtracting/issues/4

burger also has the same purpose

this is a big project, but if you feel like it's important, you could try to contribute to one of these efforts.

rom1504:
unless someone decides to spend some serious time on this, we're not going to solve this extractor problem at once

so my suggestion is instead to document what needs doing and improve the extractors as we go

rom1504 (approved):

PR #397, opened 2021-04-18
Add legacy.json 1.12->1.13 mappings

Karang:


rom1504:
can you `JSON.stringify(a, null, 2)` the file ?

rom1504:
and add a note there https://github.com/PrismarineJS/minecraft-data#data that this file is a mapping from block:metadata id from 1.12 to name in 1.13 ?

rom1504 (approved):

PR #402, opened 2021-04-25
Make add-data md file have much more info

u9g:


rom1504 (changes requested):
Can you say here that minecraft-data-auto-updater do all the automated one and should be used

u9g:
> Can you say here that minecraft-data-auto-updater do all the automated one and should be used

done

rom1504 (approved):

PR #404, opened 2021-05-19
Add bedrock edition data

extremeheat:
file name | status | how to get it | notes
-- | -- | -- | --
protocol.json | 🔵 | manual updated | wip
blocks.json | ✔ | bedrock-extractor | &nbsp;
items.json | ✔ | bedrock-extractor | &nbsp;
entities.json | ❌ |   | &nbsp;
recipes.json | ✔ |  bedrock-protocol dumper | new schema: adds furnace and other special recipes. <br/> Some recipes can have multiple outputs, also allow for recipe ingredients with damage values and specific counts.
blockCollisionShapes.json | ✔ | (from pc data) | New schema: block IDs map to an array of block state indexes, to get the block stateID, minStateId + index of array. This way different copies of the block with different state IDs can have unique collisions. 
commands.json | ❌ | bedrock-protocol + dumper | must be custom schema, current JE schema too low level
biomes.json | ✔ | bedrock-extractor (via Amulet) | 
instruments | ✔ | manual, [1](https://github.com/thebigsmileXD/libnbs/blob/master/src/xenialdan/libnbs/NBSFile.php) | &nbsp;
materials.json | ❌ | (from pc data) | &nbsp;
windows.json | ✔ | manual obtained via proxy | &nbsp;
version.json | ✔ | from bedrock-protocol | 
effects.json | ❌ | (from pc data) | &nbsp;
enchantments.json | ❌ | (from pc data) | &nbsp;
language.json | ❌ | (extract from game) | &nbsp;
particles.json | 🔵 | bedrock-protocol docs | IDs not needed, handled in protocol
blockLoot.json | ❌ | (from pc data) | &nbsp;
entityLoot.json | ❌ | (from pc data) | &nbsp;
mapIcons.json | 🔵 | ? | ?


rom1504:
let's finish this ?

extremeheat:
Added 1.17 blocks, items, windows and biomes. Added some notes and a progress board.

rom1504:
Do you think it would make sense to merge this even if not 100% finished and finish it later ? Would it help ?

rom1504:
yeah I think it would be good to do it sooner rather than later

rom1504:
hi, maybe let's add version.json and then merge it ? so the integration can be started in other places too

extremeheat:
added the version.json, though I don't think it's really needed since the same info is in the protocolVersions.json file.

Anyway, I updated the recipes.json schema to support a new format which supports all recipe types (furnace, stonecutting, shulker, etc). This could be used by #290 also. Since bedrock data isn't used anywhere right now I don't think it's breaking, but maybe it could be moved to a recpies2.json instead if that's an issue.

rom1504:
It's ok to change the format but then it would make everything easier to change it everywhere else too (pc version prismarine recipe etc)
It would be good if we can avoid having 2 format for the same thing

rom1504:
I looked at this, and it looks good to me
let's merge ?

rom1504 (approved):

rom1504:
ok I'm going to go ahead and merge this.
Feel free to open any PR with new changes if needed

indeed we'll probably need to adapt node-minecraft-data a bit (mostly pe -> bedrock I believe)

PR #410, opened 2021-06-09
Add support for 1.17 protocol

nickelpro:
Diff against last supported pre-release viewable [here](https://www.jstoolset.com/diff/80240b9fa326ba27)

Packet layout based on the Pre-Release documentation and Burger

PR #411, opened 2021-06-13
Update most relevant data for 1.17

Archengius:
This PR is built on top of 1.17 protocol data introduced in #410
Data has been obtained using [minecraft-data-generator](https://github.com/Archengius/minecraft-data-generator) and [minecraft-jar-extractor](https://github.com/PrismarineJS/minecraft-jar-extractor)


Archengius:
I've added material data to blocks and updated materials.json
Please keep in mind that materials in these files no longer actually represent block materials, but are rather based on tools by which blocks are destructible. 
This change has been made because Minecraft 1.17 no longer uses actual block materials for determining mining speed, and instead uses block tags and special code for certain items, as described in #412.

Still, by replacing materials concept, I've made actual tool speed modifiers match what you see in-game and what vanilla servers expect without making changes to other modules such as prismarine-block or bending data layouts. But again, it's rather hacky and materials just should not be used for determining harvest levels anymore.

rom1504:
Tests are failing

rom1504:
@u9g could you check it looks ok to you ?

Btw why can't we use the automated process for this ?

Archengius:
for what? 1.17 data?
because most of the relevant burger toppings are broken
all of that data is still generated automatically, just not through the burger

rom1504:
ah yeah, I meant https://github.com/PrismarineJS/minecraft-data-auto-updater
we could update https://github.com/PrismarineJS/minecraft-data-auto-updater/blob/master/.github/workflows/dump-pc-data.yml to use your tool then if it works better

rom1504 (approved):

rom1504:
ok I looked, it seems all good

u9g:
> @u9g could you check it looks ok to you ?
> 
> Btw why can't we use the automated process for this ?

@Archengius is right, a considerable amount of toppings broke in Burger, so the workflow fails because at its core it's mostly Burger files like blocks.json and items.json. Someone in the discord said that the person who develops Burger's computer broke, so it's not being updated currently for 1.17 release, and I don't have the faintest idea how to analyze the bytecode, or even where to find it, so it's stuck like that until either someone PRs a fix or pokechu's computer gets fixed.

About this new tool, I disagree with adding another tool, especially one that nobody uses at the moment to be a required tool to get the data for mcdata, we see how that works with PrismarineJS/McDataExtracting, where it took a while to find a fix because nobody from PrismarineJS knew how to use it except for them. At least with burger, there are more than one person who know how to use and maintain it. Of course, that brings us back to why is it not updated now. I also looked around the tool and looks like @Archengius is [suppressing deprecation warnings already][1], which I don't think is good for a tool that was just created.

[1]: https://github.com/Archengius/minecraft-data-generator/blob/master/src/main/java/me/arch/mcdatagen/util/DataGeneratorUtils.java#L18

PR #413, opened 2021-06-15
Standardise MD Names & Change PC to Java Edition, PE to Bedrock Edition.

KaffinPX:
^^


nickelpro:
The bedrock move is already covered by https://github.com/PrismarineJS/minecraft-data/pull/404, and I don't see the value in breaking several ecosystems by renaming pc -> je

KaffinPX:
ill fix it asap

15 Haz 2021 Sal 17:07 tarihinde Vito Gamberini ***@***.***>
şunu yazdı:

u9g:
> The bedrock move is already covered by #404, and I don't see the value in breaking several ecosystems by renaming pc -> je

Completely agree with this, breaks too much to ever be considered. and it's not like there are two pc minecraft types, I think how it is now is fine, and https://github.com/PrismarineJS/minecraft-data/pull/404 should fix the pe being merged intobedrock edition

PR #415, opened 2021-06-16
ci(Tag): remove unnecessary dispatch

u9g:
Removed from here, the dispatch should be in http://github.com/PrismarineJS/node-minecraft-data


rom1504 (approved):

rom1504:
this is incorrect, we do dispatch to nmd for releases

PR #419, opened 2021-06-21
Fix particleData IDs for 1.17

nickelpro:
@u9g ran into this bug. Particle IDs updated in 1.17 and I didn't update the hardcoded values in the particleData type. This PR fixes that bug

### DIFF
--- a/data/pc/1.17/protocol.json
+++ b/data/pc/1.17/protocol.json
@@ -92,7 +92,7 @@
       {
         "compareTo": "$compareTo",
         "fields": {
-          "3": [
+          "4": [
             "container",
             [
               {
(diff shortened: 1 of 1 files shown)


nickelpro:
This needs further investigation, the 1.17 particles.json disagrees with the wiki.vg/protocol description of 1.17 particles. Not sure which IDs are correct atm

nickelpro:
The wiki is out dated, I've updated the PR to use IDs from mcd's particles.json. This is ready for review and merge

nickelpro:
This PR now also includes the new optional particle data for the Vibration particle and the Dust Color Transition particle

Archengius:
Judging from the MC source code, particles.json IDs are correct indeed and represent what is actually used in the particle packet

rom1504 (approved):

PR #429, opened 2021-07-10
Fix resource_pack_send has an additional arg that was forgotten

u9g:


data/pc/1.17/protocol.json
```diff
@@ -3085,6 +3085,13 @@
             {
               "name": "forced",
               "type": "bool"
+            },
+            {
+              "name": "promptMesage",
```
rom1504, line 3090: typo

u9g: good catch, fixed

halotroop2288 (approved):

data/pc/1.17/protocol.json
```diff
@@ -3085,6 +3085,13 @@
             {
               "name": "forced",
               "type": "bool"
+            },
+            {
+              "name": "promptMessage",
+              "type": [
```
rom1504, line 3091: that's wrong, option already contains the bool

u9g: I don't understand what you mean, I was trying to say this is an optional string arg, is that not what that means?

rom1504: yes but you need to remove the bool field

rom1504: option means bool + something
so you need to remove the bool above or there is 2 bool

u9g: by bool field, do you mean this one? https://github.com/PrismarineJS/minecraft-data/blob/b3b1b88b4af060af1157478616ca24a3e5dc638b/data/pc/1.17/protocol.json#L3086-L3087 if so, that should still be there, if not, I'm still not understanding you

PR #434, opened 2021-07-16
Add 1.17.1 protocol.json

u9g:


u9g:
something I did notice whiel working on this is that in the window_click packet, mode is a varint, not a byte, and has been for a while, I don't know if it's worth changing the past protocol.jsons, but for this one I made the switch

rom1504:
what's the source of the info ? looks like wiki.vg is not up to date ?

u9g:
> what's the source of the info ? looks like wiki.vg is not up to date ?

I diff’d the minecraft versions by hand, then was going to add the data to wiki if it’s right. I’m not sure who usually adds to the wiki but its been a couple days and 1.17.1 support is already in a lot of big mc projects so would be good to atleast get experimental support here

rom1504:
ok can you add the version file and update the file in common ? (check similar Prs for details)

u9g:
> ok can you add the version file and update the file in common ? (check similar Prs for details)

Done

rom1504:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/common/versions.json

u9g:
> https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/common/versions.json

done

rom1504 (approved):

PR #437, opened 2021-07-20
Relocate smelting_format and tags to types map

nickelpro:
Having these types inlined in the protocol:

1. Is frustrating for consumers of protocol.json who expect all custom types in the type map
2. [Breaks the automatically generated docs](https://minecraft-data.prismarine.js.org/?v=1.17.1&d=protocol#toClient_undefined)

So let's move them so that they can play with all the other types, they'll be happier there anyway

### DIFF
--- a/data/pc/1.14.1/protocol.json
+++ b/data/pc/1.14.1/protocol.json
@@ -269,7 +269,45 @@
           ]
         ]
       }
-    ]
+    ],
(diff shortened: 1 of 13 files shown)


u9g (approved):
LGTM, tested ingame, works fine

rom1504 (approved):

PR #438, opened 2021-07-26
Re-extract all items.json

nickelpro:
This commit contains breaking changes and requires discussion, the purpose of this commit is two-fold:

* Fix a data-bug with `enchantCategories` inherited from turbo-invention (see https://github.com/u9g/turbo-invention/pull/1)

* Get all `items.json`s and the associated schema in-line with what's generated by the automated tooling. This causes two changes:
    * Removes the `durability` field, it is always either `null` or a duplicate of `maxDurability`, and turbo-invention only adds `maxDurability`
    * `fixedWith` -> `repairWith`, the data and schema disagreed here, as this field was always called `repairWith` in all versions except 1.17

### DIFF
--- a/data/pc/1.10/items.json
+++ b/data/pc/1.10/items.json
@@ -1192,7 +1192,6 @@
     "displayName": "Iron Shovel",
     "stackSize": 1,
     "name": "iron_shovel",
-    "durability": 250,
     "enchantCategories": [
       "digger",
       "breakable",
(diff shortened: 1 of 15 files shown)


u9g:
I think for fixedWith, it was just a mistake because the generator for 1.17 was different

nickelpro:
And yet the schema wanted `fixedWith`, the test simply never ran for any other data version

nickelpro:
Also as an aside, I think turbo-invention should just be rolled into burger-extractor. It doesn't make sense that the code bases are separate

rom1504 (approved):

rom1504:
thanks for the fixes
I agree about moving turbo invention code to burger extractor
what do you think @u9g ?

PR #439, opened 2021-07-27
Correctly label packet_map icons z-axis, fix #169

### LINKED ISSUE #169: map data icons.y should be z according to wiki.vg
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.11/protocol.json#L1499

Probably in other versions as well

nickelpro:
It baffles me this bug has survived every single protocol update _after being reported_, but here we are.


u9g:
can you regression test this with wherever it’s used, to be honest not sure where it is used tho

nickelpro:
I would be amazed if it's used anywhere, it's visual display data

u9g (approved):
LGTM

rom1504 (approved):

PR #440, opened 2021-07-28
Make items.json more goodly for great honor of Prismarine Project and Community

### LINKED ISSUE #199: Missing item (id:453) in 1.12
Knowledge Book (id: 453) seems to be missing in 1.12, 1.12.1, 1.12.2.
https://minecraft.gamepedia.com/Knowledge_Book

nickelpro:
This started as an attempt to fix https://github.com/PrismarineJS/minecraft-data/issues/199 and got out of control

Stuff this PR does to items.json:

* Removes non-items across all versions:
  * Air (1.13+ this is typically the only removed "item")
  * Plant stems
  * Liquids
  * Piston Heads
  * Double slabs
  * Fire
  * Redstone Wire
  * Burning Furnace
  * Placed signs
  * You get the idea
  
* Removes a bunch of items from 1.7 that aren't in 1.7

* Add missing items to 1.8+:
  * Magma Block
  * Nether Wart Block
  * Red Nether Block
  * Bone Block
  * Structure Void

* Add Knowledge Book to 1.12+ (Fixes #199)

* Fix some random stacksize bugs (ex. 1.7 had doors as stacksize 1)

* Fixes some display name capitalizations ("mall Disc" -> "Mall Disc")

* A truly laughable amount of noise from burger-extractor and turbo-inventor re-ordering fields. It's my hope that this is a one-time cost and future runs through the data won't induce so much noise. Post-flattening data versions already mostly used this field order, so now all `items.json` do.

**Now the bad news:**
This commit _doesn't_ fix the lack of "variations" provided by pre-flattening data. Ie. All `stone` still looks the same in `items.json`. We provide variations inconsistently for a handful of items, but a complete solution will have to be done in a followup

### DIFF
--- a/data/pc/1.10/items.json
+++ b/data/pc/1.10/items.json
@@ -1,11 +1,5 @@
 [
   {
-    "id": 0,
-    "displayName": "Air",
-    "name": "air",
-    "stackSize": 64
-  },
-  {
     "id": 1,
     "displayName": "Stone",
     "name": "stone",
@@ -31,7 +25,7 @@
   },
   {
     "id": 5,
-    "displayName": "Wood Planks",
+    "displayName": "Wooden Planks",
     "name": "planks",
     "stackSize": 64
   },
@@ -45,30 +39,6 @@
     "id": 7,
     "displayName": "Bedrock",
     "name": "bedrock",
-    "stackSize": 64
-  },
-  {
-    "id": 8,
-    "displayName": "Water",
-    "name": "flowing_water",
-    "stackSize": 64
-  },
-  {
-    "id": 9,
-    "displayName": "Stationary Water",
-    "name": "water",
-    "stackSize": 64
-  },
-  {
-    "id": 10,
-    "displayName": "Lava",
-    "name": "flowing_lava",
-    "stackSize": 64
-  },
-  {
-    "id": 11,
-    "displayName": "Stationary Lava",
-    "name": "lava",
     "stackSize": 64
   },
   {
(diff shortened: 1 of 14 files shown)


rom1504:
are the items you removed actually blocks ?

nickelpro:
Yes, but not blocks that can be obtained as items. They're still available in blocks.json. This is consistent with the behavior of post-flattening items.json, where we've never included non-item blocks

nickelpro:
For example, right now in pre-flattening items.json `sign` appears three times. Once as an item, once in it's free-standing block form, and again as it's wall-mounted block form. This PR removes the latter two from items.json

rom1504 (approved):

rom1504:
what tool did you use to do this change ?

nickelpro:
Burger and burger-extractor get you 70% of the way there, I have a PR open on burger-extractor that fixes minor displayName bugs (https://github.com/PrismarineJS/burger-extractor/pull/19). There's a bug in Burger that affects pre-flattening stack size which I'm tracking here (https://github.com/Pokechu22/Burger/issues/28) and fixed manually for the purpose of this commit.

Pre-flattening "variations" are not supported by any toolchain currently, what's present in this commit was carried over from the data that was already manually entered into MCD. I'm unsure if this is something Burger can support, but I'll explore that more after the Burger re-factor.

nickelpro:
Oh and for 1.7 and below everything is scuffed, Burger's identify topping is broken for versions that old so you're completely on your own and all data needs to be manhandled

PR #442, opened 2021-08-01
Fix 1.7 open_window when opening horse inventory

Beaness:
https://wiki.vg/index.php?title=Protocol&oldid=6003#Open_Window
The current switch is incorrect and is only for 1.8+. 1.7 still uses the old window ids.


nickelpro (approved):
Correct per protocol docs, LGTM

u9g (approved):
LGTM

PR #445, opened 2021-08-09
Add bedrock protocol data

extremeheat:
Adds bedrock-protocol's protocol YML data


data/dataPaths.json
```diff
@@ -1010,14 +1010,21 @@
...
-      "protocol": "bedrock/1.16.210"
+      "protocol": "bedrock/1.16.210",
+      "steve": "bedrock/1.16.201",
+      "steveSkin": "bedrock/1.16.201",
+      "steveGeometry": "bedrock/1.16.201"
     },
     "1.16.220": {
       "blockStates": "bedrock/1.16.220",
       "steve": "bedrock/1.16.201",
+      "steveSkin": "bedrock/1.16.201",
```
rom1504, line 1027: should we really have this file in mcdata ? what is it used for ?

if we want to have it, then we need to change node-minecraft-data to support non json files

data/bedrock/latest/proto.yml
```diff
@@ -0,0 +1,3171 @@
+# Created from MiNET and gophertunnel docs
```
rom1504, line 1: why do we want this to be only in a latest folder ?
isn't documentation for older versions useful too ?

data/dataPaths.json
extremeheat: So these are skin files used by the bedrock-protocol client when sending the login packet https://github.com/PrismarineJS/bedrock-protocol/blob/master/src/handshake/login.js#L10. So I think it would be possible to put it all into one big JSON with base64 if necessary

rom1504: yeah that would be more convenient

extremeheat: Done, tests are failing now with the removal but should be fixed on node-minecraft-data release

data/bedrock/latest/proto.yml
rom1504: what do you think about this @extremeheat ?

extremeheat: On Bedrock you can’t really downgrade aside from manually installing APKs or things like that.

So how about whenever a new version comes out we copy the YML then put a copy of it into the appropriate version folder? That way the docs/generated HTML for the old version can still be updated later on if needed. Also that way the latest file is still easy to diff/git blame and see where the protocol changes happen over time. 

Otherwise if that’s an issue we can do the copy thing as usual

rom1504: what's the difference between what you are proposing and the usual copy thing ?
the latest yaml file would be in the latest version we support

I think we need to decide to support multi version or not. If we don't then we should only have one latest dir and that's it.
I think there is value to support multi version, and if we do so, I don't see why not following the same structure as the rest of mcdata.

One argument to support multi version is to allow for a smoother transition period when updating libs to work for the new versions (by just adding support for the new version instead of breaking changes)

extremeheat: Oh yeah, I’m not arguing against multi version support. A lot of 3rd party things work on older versions and somethings don’t update so I think it’s important to support old versions. It’s just a matter of deciding whether to make it first class support or not, which I think it already is in a way (we do tests and all). 

Anyway, the benefit I see with having a linear “latest” file is that it’s just easier to track changes to the protocol, for example when new versions come out in PR diffs. So to see where a packet or field changed in the past, it’s simple to run git blame on it. What I was mentioning was just a compromise of keeping that history like a wiki and also being able to update docs for old versions by keeping a copy of the YAML for the old versions. But if this is a maintenance burden we can do the usual duplicate + write on top approach

rom1504: ah I see your point.
Yeah that's not a bad idea, but then I would just like to avoid the situation where latest.yml is different from $latest/protocol.yml 
So let's add a test that check the file are exactly the same and I'm ok with that solution:
* a latest dir with proto.yml and types.yml
* copy of these files in all versions
* proto.yml and types.yml of latest.yml folder should be exactly equal (exact diff) with the types.yml and latest.yml of the latest version

extremeheat: I think the having/maintaining a duplicate copy can just be avoided, for example when 1.18 comes out, the current latest files (1.17) are only then copied to the 1.17 dir. Then the 1.18 stuff is written ontop of latest. So doing it that way would avoid having to manage two copies of the same file.

rom1504: I think that's likely to make it confusing for users.

we've been doing the no latest thing since the beginning for mcpc and it worked ok

do you expect there will be many more versions for bedrock than mcpc ?

data/dataPaths.json
```diff
@@ -1043,10 +1047,9 @@
       "instruments": "bedrock/1.17.0",
       "protocol": "bedrock/1.17.10",
       "windows": "bedrock/1.16.201",
-      "blocksB2J": "bedrock/1.17.10",
-      "blocksJ2B": "bedrock/1.17.10",
       "steve": "bedrock/1.16.201",
-      "steveGeometry": "bedrock/1.16.201"
+      "blocksB2J": "bedrock/1.17.10",
```
rom1504, line 1051: Imo we should also expose the yml files here.

data/bedrock/latest/proto.yml
rom1504: Ok let's try your way then:
* We have a latest.yml and type.yml for the latest
* Copied for all other versions
* Data path points to it to expose it

Does that sound ok to you ?

data/dataPaths.json
```diff
@@ -1041,12 +1050,17 @@
...
+      "effects": "pc/1.17",
       "protocol": "bedrock/1.17.10",
       "windows": "bedrock/1.16.201",
+      "steve": "bedrock/1.16.201",
       "blocksB2J": "bedrock/1.17.10",
       "blocksJ2B": "bedrock/1.17.10",
-      "steve": "bedrock/1.16.201",
-      "steveGeometry": "bedrock/1.16.201"
+      "protoDoc": "bedrock/1.17.10",
+      "protocolYml": "bedrock/latest",
```
rom1504, line 1062: What is proto doc ?

rom1504:
I'm thinking it could be a good idea to add yml support to node-minecraft-data before merging this.

data/dataPaths.json
extremeheat: Was fixed in another commit

extremeheat:
yeah, opened https://github.com/PrismarineJS/node-minecraft-data/pull/110 which load non-JSON files as a URL

rom1504:
is this ready ? looks ok to me

PR #454, opened 2021-08-15
Add bedrock data to pages

extremeheat:
The bedrock protocol data is generated as static HTML, so it should load instantly. Right now the bundle.js is 60MB so this should probably be addressed.

ported from https://github.com/PrismarineJS/bedrock-protocol/pull/98


loadProtocol.js
```diff
@@ -293,9 +293,14 @@ function eqs (compareTo, k) {
 }
 
 function loadProtocol (version) {
-  const data = require('minecraft-data')(version).protocol
-  const comments = require('minecraft-data')(version).protocolComments
-  $j('#protocolTable').html(protocolToString(data, comments))
+  if (version.startsWith('bedrock')) {
+    const [, v] = version.split('_')
+    $j('#protocolTable').html(`<iframe src="protocol/bedrock/${v}" frameborder="0" style='width: 100%; height: 80vh;'><a href="protocol/bedrock/${v}">Click here</a></iframe>`)
```
rom1504, line 298: Do we really need an iframe ?

extremeheat: The html could be loaded into the page with jquery, however the styles used in the html file which is self contained could collide with the page styles so I used an iframe here.

rom1504: why can't we use browserify like the rest of this package ?

rom1504: ah it's this compiler thing from protodef yaml hm

rom1504: yeah well still... there has to be a better way than an iframe...

rom1504:
so actually about the iframe I don't care that much as long as it works
I tried it `npm run prod-start` and I don't see any bedrock link anywhere
should it work like this ?

loadProtocol.js
extremeheat: I took a look at removing the header and putting in the css here, but the issue is that it uses bootstrap and without it it looks a lot uglier IMO.

extremeheat:
Fixed

rom1504:
ah I just realized that this PR is targeting the wrong branch
this should be targeting the *web* branch, not gh-pages

extremeheat:
Updated

rom1504:
so I tested it and it looks really good
one issue is there is no scroll bar in the protocol page

there are probably several ways to fix it

one could be to use a web component (https://lit-element.polymer-project.org/guide) to keep the css encapsulated while not using an iframe

extremeheat:
Ah, probably an issue with the styling. Wouldn't adding lit element add extra tooling?

I'm not really familiar with web components but it [seems like it could work yeah](https://sebastiandedeyne.com/embed-a-web-page-with-a-web-component-and-the-shadow-dom/) with a shadow DOM, I'll take a look.

rom1504:
no lit element is just about adding a dep

extremeheat:
OK, I've switched it to use a web component. Since the shadow DOM doesn't get the same events as the parent, clicking on the anchor links didn't do anything (no scrolling) so I wrote a simple handler to proxy over the events.

rom1504:
there's some lint issue

extremeheat:
fixed

rom1504 (approved):

rom1504:
thanks

I noticed the anchors do not work when using a link with anchors
not sure why

extremeheat:
Ah yeah, I think that can be fixed by calling `scrollCallback` in the connectedCallback

PR #455, opened 2021-08-15
protocol packet field type mismatch fix

Gnog3:


rom1504:
Did you check in wiki.vg it's true for all these versions ?

Gnog3:
yes

Issue #456, opened 2021-08-15
Phantom u32 usage for packet_crafting_book_data


kvoli:
```
data/pc/1.12-pre4/protocol.json
data/pc/17w15a/protocol.json
data/pc/17w15b/protocol.json
```

Use the type `u32` without having it declared `native` in the types header. In all other protocol versions this is mapped to `varint`.

This slightly confuses automatic generators that look at native fields.

PR #457, opened 2021-08-15
remap u32 to varint in protocol.json

kvoli:
#456 

remove u32 and replace with varint, per every other protocol version.

### DIFF
--- a/data/pc/1.12-pre4/protocol.json
+++ b/data/pc/1.12-pre4/protocol.json
@@ -3676,7 +3676,7 @@
           [
             {
               "name": "type",
-              "type": "u32"
+              "type": "varint"
             },
             {
               "anon":true,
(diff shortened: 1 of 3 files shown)


nickelpro:
Needs to be double checked with Burger that these snapshots actually don't use u32 here

rom1504:
I don't think this is correct

just add u32 as a native

kvoli:
So all the versions before and after which are varint for that packet, are wrong?

nickelpro:
No, likely it only changed for these versions

rom1504:
> So all the versions before and after which are varint for that packet, are wrong?

looks like this packet was created in the versions in this PR. Can you find any previous version that contains this packet ?

nickelpro:
Looking at the decompile for 17w18a, this was an i32. The correct answer is to just switch to i32 which will fix the issue

kvoli:
Updated with i32 @nickelpro @rom1504

nickelpro (approved):
LGTM

rom1504 (approved):

Issue #463, opened 2021-09-01
RFC: Tag Schema


nevercast:
## Summary 

I would like to open the discussion on Tags. Burger already supports tag data, so I would like to reach a conclusion on the schema for minecraft-data. Implementing "Tags" fully probably involves recipies and other aspects of minecraft-data and probably has a rolling effect in other prismarine-* libraries, and mineflayer itself. For now, let's just think schema; or more accurately the data files and we write the schema files after.

The tags data in the minecraft jar consists of these directories: 
```
data/minecraft/tags/blocks
data/minecraft/tags/entity_types
data/minecraft/tags/fluids
data/minecraft/tags/game_events
data/minecraft/tags/items
```
Burger provides all of these.

It's my opinion that we should have a new schema for tags itself for tag -> object lookups, as well as modifying appropriate other schema for object -> tags lookups. 

## Blocks

`/blocks` is straight forward and maps to the blocks schema. I believe `blocks.json` should be modified to include tags as follows:

```js
{
  "id": 1,
  "name": "stone",
  ...,
  "tags": [
    "base_stone_overworld",
    "dripstone_replaceable_blocks",
    "lush_ground_replaceable",
    "mineable/pickaxe",
    "moss_replaceable",
    "stone_ore_replaceables"    
  ],
  ...
}
```

## Entity Types
`/entity_types` should map directly to the entities schema:

```js
{
  "id": 2,
  "name": "arrow",
  "tags": [
    "arrows"
  ]
}
```

## Fluids 
`/fluids` is not something we treat differently to blocks in minecraft-data. Water and Lava should be handled as if it were blocks (above).

## Game Events
`/game_events` is not something that I've seen mapped anywhere in minecraft-data and can be safely ignored.

## Items 
`/items` should map to the existing items schema and items.json

```js
{
    "id": 101,
    "displayName": "Oak Log",
    "name": "oak_log",
    "stackSize": 64,
    "tags": [
      "logs",
      "logs_that_burn",
      "oak_logs"
    ]
  },
```

## Tags
There should be a new file, `tags.json` with a new `tags_schema.json` that provides mappings from tag names to items, blocks and entities

Since I think it's improbable that we would ever want a tag to return a mixed array of Blocks, Items and Entities it seems obvious that the tags data should be separated by the object type such that we have "blocks", "items", and "entities" types. The data may appear as follows:

```js
{ 
  "blocks": {
    "base_stone_overworld": [
      "stone",
      "granite",
[...]

PR #464, opened 2021-09-01
Tag schemas

nevercast:
Added Schema files for blocks.json, items.json, entities.json and a new tags.json. This matches the proposal outlined here: https://github.com/PrismarineJS/minecraft-data/issues/463 

On blocks, items and entities the "tags" property is optional. However if it's specified, a minimum of one tag is required. This is to prevent accidentally adding empty tags. If there are no tags for an object, do not put a tags property on that object.

The tags have an enforced uniqueness to prevent duplicates. They do not however enforce any particular naming scheme. While not required, it could be beneficial to prevent a leading `#` if we intend on expanding nested tag references into their final values, something that I would suggest to simplify query complexity at the cost of a more complicated extractor.

As for the tags.json, the keys "blocks", "items" and "entities" are required, but there is no constraint that they are populated. This is consistent with requesting an incorrect block name and it not being present. The tag collections on each block, item or entity must have a minimum of one tag if specified, and uniqueness is enforced. This is consistent with the other files, and an entire definition should be omitted if the tag collection is empty. 

This schema requires at least Draft 4 to be parsable. I would have liked to use `contains` instead of `items`, but it's more modern and the other schema files did not use it.

### DIFF
--- a/schemas/blocks_schema.json
+++ b/schemas/blocks_schema.json
@@ -202,6 +202,13 @@
           "null"
         ],
         "minimum": -1
+      },
+      "tags": {
+        "description": "Block tags",
+        "type": "array",
+        "uniqueItems": true,
+        "minItems": 1,
+        "items": { "type": "string" }
       }
     },
     "required": [
--- a/schemas/entities_schema.json
+++ b/schemas/entities_schema.json
@@ -46,6 +46,13 @@
       "category": {
         "description": "The category of an entity : a semantic category",
         "type": "string"
+      },
+      "tags": {
+        "description": "Entity tags",
+        "type": "array",
+        "uniqueItems": true,
+        "minItems": 1,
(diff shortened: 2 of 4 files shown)


nevercast:
Can this be merged? Anything blocking it?

rom1504:
hi, thanks for the PR

I think it would make sense if you could try to generate the data for this, it's not impossible you'll find you will need to adapt the schema once you have the data

nevercast:
> hi, thanks for the PR
> 
> I think it would make sense if you could try to generate the data for this, it's not impossible you'll find you will need to adapt the schema once you have the data

Thanks for triaging this, I will begin the next step of generating data for 1.17.1 and we can go from there.

rom1504:
old, please re-open if you want to finish it

PR #465, opened 2021-09-24
bedrock: Add skin data paths for older versions

extremeheat:


rom1504:
feel free to do a release pr here

PR #466, opened 2021-09-24
Release 2.93.1

extremeheat:


rom1504:
2.93.1 is not right for 2 reasons:
1. this is a new feature so it should be 2.94.0
2. there is already a 2.93.1 in nmd

rom1504:
I pushed 2.94.0

PR #467, opened 2021-10-03
Bedrock 1.17.30 protocol data

extremeheat:
Preliminary support for bedrock 1.17.30 with just protocol data


rom1504:
Can you add the version to the readme ?

Issue #470, opened 2021-11-08
Parameter naming inconsistency


MarsRon:
There are two different spellings for parameter, ParamAter and ParamEter, throughout the library. Searching the [documentation](https://minecraft-data.prismarine.js.org/?v=bedrock_1.17.10&d=protocol) shows that the API uses paramAters, but some descriptions/notes use paramEters.

rom1504:
can you open a PR to fix it ?

PR #471, opened 2021-11-08
Fix parameter naming inconsistency (#470)

MarsRon:


nickelpro:
@extremeheat There was a reason this wasn't fixed before but ai don't remember what it was

u9g:
> @extremeheat There was a reason this wasn't fixed before but I don't remember what it was

its a very large change including being breaking

rom1504:
it was blocked by not moving to mcdata (see https://github.com/PrismarineJS/bedrock-protocol/issues/135 )
lgtm, what do you think @extremeheat ?

extremeheat:
It seems to be passing so that’s good, otherwise nothing significant seems to have changed. So far a few people have noticed it, I’m not sure how much of that is through active use vs just good attention to grammar. I’m going to assume it’s mostly the latter, so LGTM.

PR #477, opened 2021-12-09
1.18 protocol

nickelpro:
This includes two "Add 1.18 protocol" commits which should be squashed, and are provided purely so that the difference between 1.17 and 1.18 can be easily diffed.

Also includes a fix for 1.17 vibration particles that was discovered while investigating particle changes in 1.18

### DIFF
--- a/data/pc/1.18/protocol.json
+++ b/data/pc/1.18/protocol.json
@@ -3,14 +3,16 @@
     "varint": "native",
     "optvarint": "varint",
     "pstring": "native",
-    "u16": "native",
-    "u8": "native",
-    "i64": "native",
     "buffer": "native",
-    "i32": "native",
+    "u8": "native",
+    "u16": "native",
+    "u32": "native",
+    "u64": "native",
     "i8": "native",
-    "bool": "native",
     "i16": "native",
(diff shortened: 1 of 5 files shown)


data/pc/1.18/protocol.json
```diff
@@ -4274,6 +4360,15 @@
             }
           ]
         ],
+        "packet_simulation_distance": [
+          "container",
+          [
+            {
+              "name": "dist",
```
rom1504, line 4367: why not simulationDistance ?

rom1504 (changes requested):
https://wiki.vg/index.php?title=Pre-release_protocol&oldid=17160

missing simulation distance in join game
type field in block entity data

data/pc/1.18/protocol.json
nickelpro: Seems redundant and long winded, the packet gives that context

```js
client.on("simulation_distance", (packet) => {
  console.log(packet.dist);
})
```

rom1504: ok but let's at least have a full word: distance

PR #486, opened 2021-12-17
add attribute support

U5B:
normalizes names of attributes across pc versions


nickelpro:
We typically vendor a complete copy of the file for each major version. Other than that looks good

rom1504:
yeah indeed copy it for each version
btw are you sure it's the same in each version?

U5B:
> yeah indeed copy it for each version btw are you sure it's the same in each version

Attributes are not the same between versions
If you want, I could only include attributes that were in each version.

The naming schema changed in 1.16 when they added the 'minecraft:' prefix.

rom1504:
ok then we definitely need to have the correct attributes in each version

U5B:
Some versions shared the same attributes
(e.g 1.7 and 1.8, 1.9 through 1.11, 1.12, 1.13-1.15.2, 1.16-1.17)

nickelpro:
> Some versions shared the same attributes
(e.g 1.7-1.8, 1.9-1.11, 1.12-1.15.2, 1.16-1.17)

If we don't have a version's attributes that's fine, but dataPaths.json shouldn't point to incorrect info for a version so that needs fixing. Don't add an attributes key to dataPaths for versions we don't have data on. For major versions that have identical attributes, make a complete copy of the file for the base major version and then all minor versions should reference that single major version. Ex. 1.17 and 1.17.1 can both reference 1.17, but 1.18 should have its own copy even if its identical

U5B:
> > Some versions shared the same attributes
> > (e.g 1.7-1.8, 1.9-1.11, 1.12-1.15.2, 1.16-1.17)
> 
> If we don't have a version's attributes that's fine, but dataPaths.json shouldn't point to incorrect info for a version so that needs fixing. Don't add an attributes key to dataPaths for versions we don't have data on. For major versions that have identical attributes, make a complete copy of the file for the base major version and then all minor versions should reference that single major version. Ex. 1.17 and 1.17.1 can both reference 1.17, but 1.18 should have its own copy even if its identical

Thank you for your input.

U5B:
I couldn't get MCP to work.
However I got DecompilerMC to work and got attributes from the code for 1.14 to 1.17.
1.13 and below was just testing putting on armor and giving myself luck

data/pc/1.10/attributes.json
```diff
@@ -0,0 +1,67 @@
+[
+  {
+    "id": 1,
+    "name": "maxHealth",
```
rom1504, line 4: what's the difference between name and display name ? where is each used ?
do we want the default/min/max as well https://wiki.vg/Protocol#Entity_Properties ?
what is the id ?

in general: how will the data be used, is there enough info in that file to use it ?

U5B: name is the name that prismarine-physics will call.
displayName is the name that minecraft uses.

Basically displayName can vary between versions but I want name to be consistant between versions.

```js
function getMojangAttribute (name) {
  return mcData.mojangAttribute(name) // returns displayName
}

getMojangAttribute('movementSpeed')
// returns generic.movementSpeed for pre 1.16
// returns minecraft:generic.movement_speed for 1.16+
```

U5B: The id is what I copied from how the effects are used. I can remove it if that is needed.

rom1504: The protocol does not know names
I think maybe it knows ids. Is it these IDs?
That's what would be interesting to check

U5B: Protocol gives mojangNames

nickelpro:
The usefulness of the `id` field is questionable, since it's not mapped to anything actually in use in MC (unless I'm missing something?). I don't know if `mojangName` is the best name for that field either, MC calls these variously ["resource identifiers" or "namespaced IDs"](https://minecraft.fandom.com/wiki/Resource_location). I think we would be fine just calling that field `resource` or something to that effect, to try to stick close to what MC developers expect.

But this is nitpicking, this commit is in a good place already, good work so far. @rom1504 How do you feel about these suggestions?

rom1504:
I think "name" is the right name considering other things in Minecraft data
Let's remove id since it's confusing


What about min/max/default ?

nickelpro:
> I think "name" is the right name considering other things in Minecraft data Let's remove id since it's confusing

`name` is already used for the identifier we're keeping constant for any given attribute. The question is what to call the identifier that Mojang keeps changing, which we won't use internal to prismarine but is necessary for serializing/deserializing the attribute modifier packets.

rom1504:
did mojang name actually change between 2 versions ?

nickelpro:
Yes, it sometimes changes and it gained the `minecraft:` prefix between 1.15 and 1.16

rom1504:
ok, then let's just remove id and keep the rest as is

rom1504:
Next step is to add support in node-minecraft-data then use it somehow in mineflayer

U5B:
> Next step is to add support in node-minecraft-data then use it somehow in mineflayer

Should I make a release PR for minecraft-data so I can start work in node-minecraft-data?

PR #487, opened 2021-12-21
Add data from 1.18

rom1504bot:
Auto-generated by a GH workflow


rom1504:
1) audit recipes pc 1.18
       pickaxe not upside-down:

we need that fix

nickelpro:
Just need to run the audit with the "flip the recipe" code uncommented, then commit the changes. I'll open a bug on Burger extractor to fix this in the future.

PR #490, opened 2021-12-23
Create pre_flattening_texturepack_mappings.json

u9g:
Made from https://github.com/extremeheat/extracted_minecraft_data/blob/client1.18.1/client/net/minecraft/client/resources/LegacyPackResourcesAdapter.java#L721


rom1504:
Can you say what it is in https://github.com/PrismarineJS/minecraft-data#data ?
(So we don't forget in a month what this file is)

PR #491, opened 2021-12-28
Create legacy_texturepack_mappings.json

u9g:
Added from `client/net/minecraft/client/resources/LegacyPackResourcesAdapter.java` in the vanilla client of 1.18.1


rom1504:
actually both this file and the one from the previous PR do not make sense in minecraft data since minecraft data does nothing with textures
they should both be in minecraft assets
can you move them there ?

u9g:
> actually both this file and the one from the previous PR do not make sense in minecraft data since minecraft data does nothing with textures they should both be in minecraft assets can you move them there ?

Yup

PR #504, opened 2022-01-23
Add cake outshapes

nickelpro:
As everyone knows, when you bake a cake you are left holding three metal buckets. Bakeries are overwhelmed by piles of metal buckets.


SaubereSache:
https://github.com/PrismarineJS/mineflayer/pull/2418

PR #507, opened 2022-01-28
bedrock: version data parity, block state and protocol fixes

extremeheat:
* make bedrock versions.json file identical to protocolVersions.json file
* fix command block update packet by working around https://github.com/ProtoDef-io/node-protodef/issues/132 and using `false` instead of `default` in switch condition
* fix animation packet boat field
* move 1.18 block data to 1.17.40 to reflect no block changes happened in 1.18
* add a dataPaths audit


rom1504:
What is release type ?

extremeheat:
it's copied from the protocolVersions.json schema, it refers to if the version is a full release/beta/snapshot version

rom1504:
if you want to use the version files, you need to update https://github.com/PrismarineJS/minecraft-data/blob/master/data/dataPaths.json for bedrock

rom1504:
btw we should have a test checking that dataPaths contains everything, otherwise it causes weird issues

extremeheat:
added with a dataPaths audit test

PR #519, opened 2022-03-01
Add issue templates so we stop screwing up new versions

nickelpro:
If checklists work for airline pilots they can work for us.


u9g:
Thanks :)

PR #526, opened 2022-03-09
Add prismarine-item features

u9g:


rom1504:
ci failed

data/pc/common/features.json
```diff
@@ -359,5 +359,35 @@
...
+    "description": "item serialization in older versions uses blockId field to show nullability by setting blockId to -1",
+    "versions": ["1.8_major", "1.12_major"]
+  },
+  {
+    "name": "preFlattening",
+    "description": "in older versions, enchant ids were used to store enchants in nbt",
+    "versions": ["1.8_major", "1.12_major"]
+  },
+  {
+    "name": "theFlattening",
```
rom1504, line 384: pick a feature name related to what this does (like enchant)

u9g: the flattening effected a lot of things, too many to point to one: https://minecraft.fandom.com/wiki/Java_Edition_1.13/Flattening

rom1504: yes, I still think we should have one feature for one thing
features shouldn't be used only to identify a version range

btw I don't think this specific feature is that related with theFlattening (unlike blockStateId feature for example)

PR #528, opened 2022-03-10
Fix effect naming in 1.17

u9g:


nickelpro:
Based on what?

rom1504:
Mojang code

PR #532, opened 2022-03-18
Modify bedrock/java edition block map schema for easier parsing

extremeheat:
Modifies the bedrock-to-java edition block mapping data schema to be friendlier to parsing, lessening the need to do string manipulation

```json
"minecraft:cobblestone_wall[wall_block_type=cobblestone,wall_connection_type_east=none,wall_connection_type_north=none,wall_connection_type_south=none,wall_connection_type_west=none,wall_post_bit=1]": "minecraft:cobblestone_wall[east=none,north=none,south=none,up=true,waterlogged=false,west=none]",
```
=>

```json
  "cobblestone_wall": {
    "wall_block_type=cobblestone,wall_connection_type_east=none,wall_connection_type_north=none,wall_connection_type_south=none,wall_connection_type_west=none,wall_post_bit=1": [
      "cobblestone_wall",
      "east=none,north=none,south=none,up=true,waterlogged=false,west=none"
    ],
```


rom1504:
What do you think about going directly for a decent schema then?
Something that looks like all the other schema of Minecraft data: a list of object with key property name

Could this all be flat ?

The schema is far from being explicit currently

extremeheat:
Yeah, that could be done. Codewise though keeping it in a tree structure makes it a bit easier to perform lookups

```js
let block = Block.fromStateId(n)
const stateKey = Object.entries(block.getProperties()).map(e => e.join('=')).join(',')
let [name, newStateKey] = mapping[block.name][stateKey]
let newStates = Object.fromEntries(newState.split(',').map(e => e.split('=')))
let newBlock = Block.fromProperties(name, newStates)
```

vs
```json
  "cobblestone_wall": [
    [
      { 
        "wall_block_type": "cobblestone",
        "wall_connection_type_east": "none",
        "wall_connection_type_north": "none",
        "wall_connection_type_south": "none",
        "wall_connection_type_west": "none",
        "wall_post_bit": "1"
      },
      "cobblestone_wall",
      {"east": "none", "north": "none", "south": "none", "up": "true", "waterlogged": "false", "west": "none"}
    ],
    [
      // ...
    ]
  ]
```
```js
let block = Block.fromStateId(n)
let originalStates = Object.entries(block.getProperties())
let mappedStates = mapping[block.name]
let newBlock
for (const mappedState of mappedStates) {
  let states = mappedState[0]
  let mapStates = mappedState[2]
  if (originalStates.every(([k, v]) => mapStates[k] == v)) { // loose compare
    newBlock = Block.fromProperties(mappedState[1], newStates)
  }
}
```

one way to simplify but keep the tree would be to make the mapped blockstates value an object
```json
  "cobblestone_wall": {
    "wall_block_type=cobblestone,wall_connection_type_east=none,wall_connection_type_north=none,wall_connection_type_south=none,wall_connection_type_west=none,wall_post_bit=1": [
      "cobblestone_wall",
      {"east": "none", "north": "none", "south": "none", "up": "true", "waterlogged": "false", "west": "none"}
    ],
```
```js
let block = Block.fromStateId(n)
const stateKey = Object.entries(block.getProperties()).map(e => e.join('=')).join(',')
let [name, newStates] = mapping[block.name][stateKey]
let newBlock = Block.fromProperties(name, newStates)
```

rom1504:
I'd suggest to do the same as for other minecraft data things: we have a more flat structure with explicit names here and we no indexing
Then in node-minecraft-data we add any indexes that can be helpful to users.

However I believe this makes sense only if we remove completely the tree structure from the json, something in the middle is not an huge improvement

rom1504:
please reopen if you want to finish it

Sertonix:
It would make sense to directly include this data in the blocks.json or blockStates.json files.
eg. in blocks.json
```json
{
  "id": 20,
  "name": "sapling",
  "displayName": "Oak Sapling",
  "hardness": 0,
  ...
  "toJava": {
    "[age_bit=0,sapling_type=oak]": "minecraft:oak_sapling[stage=0]",
    "[age_bit=1,sapling_type=oak]": "minecraft:oak_sapling[stage=1]"
  }
}
```
this would make it easy to lookup and flat

extremeheat:
No, that will greatly increase the size of the file for no gain. Mapping data is different than most block lookups, it's mainly only needed to be done when looking up block models.

Sertonix:
Ok.
Other topic: Could in the map file be included which version is used to convert to? This might not be important in most cases but it is important when you want your code to work precise.

extremeheat:
Possibly. But since Minecraft Java and Bedrock mostly use the same versioning (major/minor) you should be able to rely on this to assume the mapping. I don't think we track differences in patch version, the generator just maps to the latest minecraft java/bedrock version for the major.minor version.

PR #537, opened 2022-04-18
Fixed hardness & resistance of copper_ore

sefirosweb:


IceTank:
What do you base those values on? Are there other values that are the same as the copper ore ones?

sefirosweb:
I copied from: https://github.com/PrismarineJS/minecraft-data/blob/7e1e3e3baee9e492971b33773afdf90a7937c064/data/bedrock/1.17.10/blocks.json

PR #541, opened 2022-04-21
Push tags on release without external action

extremeheat:
Workaround https://github.com/actions-ecosystem/action-push-tag/issues/10


rom1504:
Can we do something like https://github.com/PrismarineJS/prismarine-template/blob/master/.github/workflows/publish.yml instead to have something coherent between packages

extremeheat:
yeah, updated

PR #551, opened 2022-05-05
Add data

rom1504bot:
Auto-generated by a GH workflow


u9g:
changes in `tints.json` -> we now can find tint of all biomes, as opposed to before where biomes we didnt find had a 0 for color

PR #580, opened 2022-06-15
Fix 1.18 blockLoot/entityLoot

Eejit43:


rom1504:
What change did you make ?

Eejit43:
> What change did you make ?

Check the changes, appears 1.17 data was used instead of 1.18 data

PR #589, opened 2022-07-18
1.19 protocol correctness

rob9315:
additions to #587

### DIFF
--- a/data/pc/1.19/protocol.json
+++ b/data/pc/1.19/protocol.json
@@ -5075,27 +5075,27 @@
             },
             {
               "name": "offset_x",
-              "type": "u8"
+              "type": "i8"
             },
             {
               "name": "offset_y",
-              "type": "u8"
+              "type": "i8"
             },
             {
               "name": "offset_z",
-              "type": "u8"
+              "type": "i8"
             },
             {
               "name": "size_x",
-              "type": "u8"
+              "type": "i8"
             },
             {
               "name": "size_y",
-              "type": "u8"
+              "type": "i8"
             },
             {
               "name": "size_z",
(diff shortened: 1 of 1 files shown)


data/pc/1.19/protocol.json
```diff
@@ -4867,15 +4867,15 @@
                             "countType": "varint"
                           }
                         ]
-                      },
-                      {
-                        "name": "signedPreview",
-                        "type": "bool"
```
u9g, line 4873: huh? https://wiki.vg/Protocol#Chat_Command

data/pc/1.19/protocol.json
```diff
@@ -5602,7 +5608,7 @@
           [
             {
               "name": "slotId",
-              "type": "i8"
```
u9g, line 5605: whats the point in this change, it's not valid to send a slot id > 8, as the hotbar is 9 slots long...

data/pc/1.19/protocol.json
u9g: oh nevermind

data/pc/1.19/protocol.json
extremeheat: Yeah, this is serializing into a stream of bits. Writing wrong size will mean the number will be read incorrectly (especially since it's big endian not little), as will further fields be in the wrong position.

rob9315:
i'll be adding the 1.19 equivalents of #591 here as well

data/pc/1.19/protocol.json
u9g: Okay sounds right

PR #591, opened 2022-07-19
some fixes 1.7 - 1.18.2+

rob9315:
this will also be relevant to 1.19

# fixes

## `play/toServer/packet_block_dig`
field `status` is a varint, not an i8
affected versions 1.9 .. 1.18.2+

## `play/toServer/packet_update_structure_block`
fields `offset_x`, `offset_y`, `offset_z`, `size_x`, `size_y` and `size_z` are i8 and not u8
`offset_{x,y,z}` is furthermore even allowed to have negative values which would have been read incorrectly as a big positive offset
affected versions: 1.13 .. 1.18.2+

## `play/toServer/packet_scoreboard_score`
field `action` is a varint, not an i8
affected versions: 1.7 .. 1.18.2+

### DIFF
--- a/data/pc/1.10-pre1/protocol.json
+++ b/data/pc/1.10-pre1/protocol.json
@@ -2490,7 +2490,7 @@
             },
             {
               "name": "action",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "scoreName",
--- a/data/pc/1.10-pre1/protocol.json
+++ b/data/pc/1.10-pre1/protocol.json
@@ -3320,7 +3320,7 @@
           [
             {
               "name": "status",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "location",
--- a/data/pc/1.10/protocol.json
+++ b/data/pc/1.10/protocol.json
@@ -2490,7 +2490,7 @@
             },
             {
               "name": "action",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "scoreName",
--- a/data/pc/1.10/protocol.json
+++ b/data/pc/1.10/protocol.json
@@ -3320,7 +3320,7 @@
           [
             {
               "name": "status",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "location",
--- a/data/pc/1.11/protocol.json
+++ b/data/pc/1.11/protocol.json
@@ -2490,7 +2490,7 @@
             },
             {
               "name": "action",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "scoreName",
--- a/data/pc/1.11/protocol.json
+++ b/data/pc/1.11/protocol.json
@@ -3325,7 +3325,7 @@
           [
             {
               "name": "status",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "location",
--- a/data/pc/1.12-pre4/protocol.json
+++ b/data/pc/1.12-pre4/protocol.json
@@ -3625,7 +3625,7 @@
           [
             {
               "name": "status",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "location",
--- a/data/pc/1.12-pre4/protocol.json
+++ b/data/pc/1.12-pre4/protocol.json
@@ -2707,7 +2707,7 @@
             },
             {
               "name": "action",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "scoreName",
--- a/data/pc/1.12.1/protocol.json
+++ b/data/pc/1.12.1/protocol.json
@@ -3615,7 +3615,7 @@
           [
             {
               "name": "status",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "location",
--- a/data/pc/1.12.1/protocol.json
+++ b/data/pc/1.12.1/protocol.json
@@ -2743,7 +2743,7 @@
             },
             {
               "name": "action",
-              "type": "i8"
+              "type": "varint"
             },
             {
               "name": "scoreName",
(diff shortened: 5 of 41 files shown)


extremeheat (approved):

rom1504:
This seems quite unlikely
Can you try it with nmp tests and validate it still works ?
Maybe even mineflayer tests

rob9315:
all of these changes have been directly pulled from https://wiki.vg, here the links for latest (i checked the old versions' pages as well)
[`play/toServer/packet_block_dig`](https://wiki.vg/Protocol#Player_Action)
[`play/toServer/packet_update_structure_block`](https://wiki.vg/Protocol#Program_Structure_Block)
[`play/toClient/packet_scoreboard_score`](https://wiki.vg/Protocol#Update_Score)

for the `i8` -> `varint` changes, this could be i8 before because of the range of values, but for correctness varint is more appropriate
for the structure block, this was an actual issue as negative values were possible in the `offset`s, yet not parsed correctly by `u8`

rob9315:
nmp tests ... work?
```sh
  2474 passing (9m)
  22 pending
  1 failing

  1) client 1.13.2
       "before all" hook in "client 1.13.2":
     Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/rob9315/Desktop/node-minecraft-protocol/test/clientTest.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

nickelpro:
These need to be a least spot checked, I'll do it when I get back from work

rob9315:
running nmp tests a second time results in no errors
```
  2478 passing (7m)
  24 pending```

**Comment by rob9315** [comment]
1.19 source:
```java
// 1.19/client/net/minecraft/network/protocol/game/ServerboundPlayerActionPacket.java
public class ServerboundPlayerActionPacket
implements Packet<ServerGamePacketListener> {
    //...
    public ServerboundPlayerActionPacket(FriendlyByteBuf friendlyByteBuf) {
        this.action = friendlyByteBuf.readEnum(Action.class);
        this.pos = friendlyByteBuf.readBlockPos();
        this.direction = Direction.from3DDataValue(friendlyByteBuf.readUnsignedByte());
        this.sequence = friendlyByteBuf.readVarInt();
    }
    //...
    @Override
    public void write(FriendlyByteBuf friendlyByteBuf) {
        friendlyByteBuf.writeEnum(this.action);
        friendlyByteBuf.writeBlockPos(this.pos);
        friendlyByteBuf.writeByte(this.direction.get3DDataValue());
        friendlyByteBuf.writeVarInt(this.sequence);
    }
    //...
}
```java
// 1.19/client/net/minecraft/network/FriendlyByteBuf.java
public class FriendlyByteBuf
extends ByteBuf {
    //...
    public <T extends Enum<T>> T readEnum(Class<T> class_) {
        return (T)((Enum[])class_.getEnumConstants())[this.readVarInt()];
    }
    //...
    public FriendlyByteBuf writeEnum(Enum<?> enum_) {
        return this.writeVarInt(enum_.ordinal());
    }
    //...
}
```

rob9315:
update_structure_block:
```java
        int n = 48;
        this.offset = new BlockPos(Mth.clamp((int)friendlyByteBuf.readByte(), -48, 48), Mth.clamp((int)friendlyByteBuf.readByte(), -48, 48), Mth.clamp((int)friendlyByteBuf.readByte(), -48, 48));
        int n2 = 48;
        this.size = new Vec3i(Mth.clamp((int)friendlyByteBuf.readByte(), 0, 48), Mth.clamp((int)friendlyByteBuf.readByte(), 0, 48), Mth.clamp((int)friendlyByteBuf.readByte(), 0, 48));
```

rob9315:
so yeah, all enums written with `writeEnum` and read with `readEnum` are encoded with varints
if they do not have values above 127, that is the exact same as `i8` which is why it worked so far

rob9315:
mineflayer tests also were successful (don't know why they wouldn't be but whatever)
```
  506 passing (34m)
```

nickelpro:
Ya I didn't doubt it was true for 1.19, the question is did read/writeEnum (and other updated fields in this PR) work that way all the way back to 1.7

rob9315:
i'll check 1.7 but the wiki always said VarInt Enum

rob9315:
you were right, 1.7 does not use varint for that (even though it essentially doesn't make a difference with the potential values for that)

rob9315:
i will go through the source to check in which version that was changed

rob9315:
it was 1.8 which used an enum and uses varint for enum encoding
```java    /**
     * Reads the raw packet data from the data stream.
     */
    public void readPacketData(PacketBuffer buf) throws IOException
    {
        this.status = (C07PacketPlayerDigging.Action)buf.readEnumValue(C07PacketPlayerDigging.Action.class);
        this.position = buf.readBlockPos();
        this.facing = EnumFacing.getFront(buf.readUnsignedByte());
    }
```

rob9315:
the other change to 1.7 is correct though, action really is an enum encoded as a varint there
```java
    /**
     * Reads the raw packet data from the data stream.
     */
    public void readPacketData(PacketBuffer buf) throws IOException
    {
        this.name = buf.readStringFromBuffer(40);
        this.action = (S3CPacketUpdateScore.Action)buf.readEnumValue(S3CPacketUpdateScore.Action.class);
        this.objective = buf.readStringFromBuffer(16);

        if (this.action != S3CPacketUpdateScore.Action.REMOVE)
        {
            this.value = buf.readVarIntFromBuffer();
        }
    }
```

rob9315:
ok, fixed. 1.7 did not yet use enums for those fields, this is now again reflected in the file by using i8 like before

PR #592, opened 2022-07-19
replace `varint` with custom native `varlong` where wiki.vg lists VarLong

rob9315:
the purpose is to mark fields which are actually `varlong`s in the minecraft protocol as such to allow for more accurate implementations in lower level languages than javascript in which the specific number type actually matters.
the PR which adds `varlong` as a custom minecraft type to nmp is [here](https://github.com/PrismarineJS/node-minecraft-protocol/pull/1018) and just aliases it to `varint`

once 1.19 is merged, these fixes will also have to be applied there

### DIFF
--- a/data/pc/21w07a/protocol.json
+++ b/data/pc/21w07a/protocol.json
@@ -1,6 +1,7 @@
 {
   "types": {
     "varint": "native",
+    "varlong": "native",
     "optvarint": "varint",
     "pstring": "native",
     "u16": "native",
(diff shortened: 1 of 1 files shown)


rom1504:
Would be nice to first add this to protodef and node-protodef

rob9315:
there is a pr for node-protodef

rob9315:
there is a pr for nmp and not protodef because i think `varlong` is more like `entityMetadataLoop` and `topBitSetTerminatedArray` in that it is specifically for minecraft and because there was some discussion on how to properly handle the standards situation.

rom1504:
I think there's not a lot of drawback to putting it in protodef though

Issue #524, opened 2022-03-08
PartialReadError: Read error for undefined : undefined


alexhenke01:
![error](https://user-images.githubusercontent.com/70412431/157147566-6dc0d73d-3a49-4751-a71c-6c6a2aa387e7.PNG)

u9g:
Error with declare_commands on 1.18.2

PR #596, opened 2022-07-20
packet_declare_commands 1.13 .. 1.18 & 1.19

rob9315:


rob9315:
fixes #524

rom1504:
Can you say what you changed?

rob9315:
i don't know if it's quite ready yet, but tseting, declare_commands has never worked
and looking at it i am gonna say that is because all bitfields' fields were in reverse order

rob9315:
no, looking at it, i'm wrong. it didn't work in 1.19 because it was changed from using strings to varints as keys and i royally broke all versions... *facepalm

rob9315:
alright, now it should be working, tested in 1.19

nickelpro:
I'm not sure this is a great approach. Those identifiers weren't "missing" from older versions, they were covered by the void default. Now you've enumerated a ton of strings that all result in... nothing

rob9315:
i can remove the one or two additional strings before 1.16 though

nickelpro:
No that makes sense, I retract my comment. I hadn't considered the error state.

rom1504:
So can you please describe the changes you made ?
It'll be helpful in the future if there's any problem with that packet

rob9315:
for 1.13 (where it was introduced) to 1.18:
- added missing identifiers, also the ones which don't encode any additional data to the switch statement
- switch statement does not have a default case anymore, as to not read into arbitrary data for unknown tags

for 1.19:
- instead of the `parser` string, the server now sends `parserId`, a `varint`
- converted `parser` from `string` to `mapper`, retaining old identifiers
- removed unused identifiers from switch statement
- added new identifiers to switch statement

details on [wiki.vg](https://wiki.vg/Command_Data#Parsers)

rob9315:
if there are any future problems with this packet, a parserId was likely added and wiki.vg should be looked at again

rom1504:
Great thanks

PR #597, opened 2022-07-20
move array back to declare_commands

rob9315:


rob9315:
i only changed like 8 lines per version

PR #598, opened 2022-07-20
replace `varint` with custom native `varlong` where wiki.vg lists VarLong

rob9315:
as the [old pr](https://github.com/PrismarineJS/minecraft-data/pull/592) was accidentally irrevertably closed, here a new one

the purpose is to mark fields which are actually varlongs in the minecraft protocol as such to allow for more accurate implementations in lower level languages than javascript in which the specific number type actually matters.
the PR which adds varlong as a custom minecraft type to nmp is [here](https://github.com/PrismarineJS/node-minecraft-protocol/pull/1018) and just aliases it to varint

### DIFF
--- a/data/pc/1.10-pre1/protocol.json
+++ b/data/pc/1.10-pre1/protocol.json
@@ -1,6 +1,7 @@
 {
   "types": {
     "varint": "native",
+    "varlong": "native",
     "pstring": "native",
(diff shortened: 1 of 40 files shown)


rom1504:
Ok let's do it, it's easy to revert if we change our mind

Fyi @nickelpro

PR #607, opened 2022-07-25
[1.19] fix `packet_map` protodef type

rob9315:


rob9315:
actually, this goes further back than 1.19, don't merge yet

PR #608, opened 2022-07-25
[1.7 - 1.8] `packet_entity_update_attributes`: `UUID` -> `uuid`

rob9315:
to bring these older versions back in line with the newer ones, the field names should be adjusted.
this caused a bug with which effects like speed were applied twice, see [prismarine-physics](https://github.com/PrismarineJS/prismarine-physics/pull/74).


nickelpro (approved):
LGTM @rom1504

PR #609, opened 2022-07-26
[1.19] clarify crypto-related field naming during login state

rob9315:
having `signature.signature` is comfusing and renaming the entire object to `crypto` seems logical
`messageSignature` doesn't convey the task of the field or how it's related to the previously received clientbound encryption_begin. `encryptedVerifyToken` clearly explains that the field contains the previously received `verifyToken` but encrypted.

### DIFF
--- a/data/pc/1.19/protocol.json
+++ b/data/pc/1.19/protocol.json
@@ -1282,7 +1282,7 @@
               "type": "string"
             },
             {
-              "name": "signature",
+              "name": "crypto",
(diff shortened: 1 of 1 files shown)


extremeheat:
"Crypto" makes it seem like this is related to encryption, don't see anything wrong with sign field as is

rom1504:
the previous names were fine, no ?

rob9315:
probably

PR #610, opened 2022-07-26
changed effects name to lowercase

FlameQuard:
this change will help users to use minecraft-assets easily and other json (eg. blocks.json....) are of same format

### DIFF
--- a/data/pc/1.19/effects.json
+++ b/data/pc/1.19/effects.json
@@ -1,199 +1,199 @@
 [
   {
     "id": 1,
-    "name": "Speed",
+    "name": "speed",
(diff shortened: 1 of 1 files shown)


nickelpro:
1) If we were going to change this we would do so for all versions not just 1.19

2) For files that come from a generator, PR the generator. In this case:
https://github.com/PrismarineJS/minecraft-data-generator-server/blob/main/src/main/java/dev/u9g/minecraftdatagenerator/generators/EffectsDataGenerator.java

Closing this PR because the approach isn't viable. If you want to discuss the issue further open an issue about it.

PR #612, opened 2022-08-07
fix #611

### LINKED ISSUE #611: blocksB2J and blocksJ2B  uses wrong syntax
The syntax bedrock uses for block states is not the same as in java edition.
eg.
`minecraft:note_block[instrument=pling,note=24,powered=false]`
should be:
`minecraft:note_block["instrument":"pling","note":24,"powered":false]`

[McWiki](https://minecraft.fandom.com/wiki/Argument_types#BlockStateCommandParam)

Sertonix:
script used:
```js
const fs = require('fs').promises
const path = require('path')

function replaceContent(path,callback) {
    return fs.readFile(path).then( buffer => callback(buffer.toString("utf8")) ).then( string => fs.writeFile( path, string, "utf8") )
}

async function allDirectories(path) {
    return (await fs.readdir(path,{withFileTypes:true})).filter( dirent => dirent.isDirectory() )
}

async function allFiles(path) {
    return (await fs.readdir(path,{withFileTypes:true})).filter( dirent => dirent.isFile() )
}

function editJSONEntries( str, callback ) {
    return JSON.stringify(Object.fromEntries(Object.entries(JSON.parse(str)).map( callback ) ),null,2)
}

(async()=>{
    (await allDirectories(__dirname)).map( async versionDirent =>
        (await allFiles(path.join(__dirname,versionDirent.name))).map( dirent =>
            /blocks(?:B2J|J2B)\.json/.test(dirent.name) &&
            replaceContent( path.join(__dirname,versionDirent.name,dirent.name), str =>
                editJSONEntries(str, entry => {
                    const i = dirent.name === "blocksB2J.json" ? 0 : 1;
                    entry[i] = entry[i].replace( /(?<=[[,])([a-z_]+)=([a-z_0-9]+)(?=]|,)/g, (...[,k,v]) => `"${k}":${ /^(?:[0-9]+|true|false)$/.test(v) ? v : `"${v}"` }` )
                    return entry;
                } )
            )
        )
    )
})()
```


Sertonix:
see #611

rom1504:
Why close

Sertonix:
I closed it because the schema might change again. If this will be the final schema I will commiit that again

PR #613, opened 2022-08-08
bedrock uses 0,1 instead of true,false

Sertonix:
script used:
````js
const fs = require('fs').promises
const path = require('path')

function replaceContent(path,callback) {
    return fs.readFile(path).then( buffer => callback(buffer.toString("utf8")) ).then( string => fs.writeFile( path, string, "utf8") )
}

async function allFiles(path) {
    return (await fs.readdir(path,{withFileTypes:true})).filter( dirent => dirent.isFile() )
}

function editJSONEntries( str, callback ) {
    return JSON.stringify(Object.fromEntries(Object.entries(JSON.parse(str)).map( callback ) ),null,2)
}

function fix(str) {
    return str.replace( /true|false/g, b => b === "true" ? 1 : 0 )
}

(async()=>{
    (await allFiles(__dirname))
        .filter( file => /blocks(?:B2J|J2B)\.json/.test(file.name) )
        .map( file => {
            replaceContent( path.join(__dirname,file.name),
                {
                    "blocksB2J.json": str => editJSONEntries(str, ([b,j]) => [fix(b),j] ),
                    "blocksJ2B.json": str => editJSONEntries(str, ([j,b]) => [j,fix(b)] ),
                }[file.name]
            )
        } )
})()
``


Sertonix:
Sorry, I need to check more what is correct. The syntax with true and false is correct in bedrock

PR #614, opened 2022-08-08
expanded B2J 1.19

Sertonix:
* added missing states to blocksB2J.json
Note: I tried to convert as many blockstates as possible which means that the block might not be the same when converted back


extremeheat:
Changing the block states is a breaking change. Where did you get that data ? Can you update the extractor ?

Sertonix:
Ok, I double checked and it was an error in mcwiki

rom1504:
will you finish this ?

Sertonix:
This is finished. I can do this for other versions later but not in this pull request.

rom1504:
@extremeheat what do you think about this pr ?

RaphiMC:
I found some invalid block state mappings regarding missing facing values.
Here is an example:
`"minecraft:gray_glazed_terracotta[facing_direction=0]": "minecraft:gray_glazed_terracotta[]",`

Java Edition has no gray_glazed_terracotta without a facing value.
Here is a list of all invalid Java Block states I found: https://pastebin.com/8BMKcxFu

rom1504:
old

PR #618, opened 2022-08-16
bedrock: fix 1.19.20 map data packet

extremeheat:
Fix https://github.com/PrismarineJS/bedrock-protocol/issues/256


stevarino:
Came to the same conclusion with https://github.com/PrismarineJS/minecraft-data/pull/619

PR #616, opened 2022-08-10
Bedrock 1.19.20 Update

stevarino:
Logs in, but there seems to be a problem with `packet_modal_form_response` and I'm not sure how to debug this.

```
  minecraft-data schemas bedrock 1.19.20
{
  "keyword": "enum",
  "dataPath": "",
  "schemaPath": "#/oneOf/0/enum",
  "params": {
    "allowedValues": [
      "native"
    ]
  },
  "message": "should be equal to one of the allowed values",        
  "schema": [
    "native"
  ],
  "parentSchema": {
    "enum": [
      "native"
    ]
  },
  "data": [
    "container",
    [
      {
        "name": "form_id",
        "type": "varint"
      },
      {
        "name": "has_response_data",
        "type": "bool"
      },
      {
        "name": "data",
        "type": [
          "switch",
          {
            "compareTo": "has_response_data",
            "fields": {
              "true": "string"
            },
            "default": "void"
          }
        ]
      },
        "type": "bool"
      },
      {
        "name": "data",
        "type": [
          "switch",
          {
            "compareTo": "has_cancel_reason",
            "fields": {
              "true": [
                "container",
                [
                  {
                    "name": "%n,0",
                    "type": "closed"
                  },
                  {
                    "name": "%n,1",
                    "type": "busy"
                  }
                ]
              ]
            },
            "default": "void"
          }
        ]
      }
    ]
  ]
}
    1) protocol.json is valid
    √ version.json is valid

  minecraft-data schemas of common data of pc
    √ protocolVersions.json is valid (59ms)
    √ features.json is valid

  minecraft-data schemas of common data of bedrock
    √ protocolVersions.json is valid
    √ features.json is valid
       protocol.json is valid:
     Error: Error at root.types.packet_modal_form_response
```

PR #622, opened 2022-08-24
1.19.21 Initial Commit

stevarino:
Looks like it is just the protocol version that changed, I can't find anything else and nothing is listed over at gophertunnel.


stevarino:
Regarding the feedback in https://github.com/PrismarineJS/minecraft-data/pull/616 - unsure how to do a release PR but I think I got everything else.

doc/history.md
```diff
@@ -1,3 +1,6 @@
+## 3.10.3
+* bedrock: add protocol data for 1.19.21 (@stevarino)
```
extremeheat, line 2: History should be updated in a separate release PR so the GH release action can be triggered

extremeheat:
A release PR is basically updating the HISTORY.md file and making a commit with the format of "Release x.x.x". The publish CI action will then be triggered. Should happen in a separate PR.

extremeheat (approved):

extremeheat:
Thanks!

PR #623, opened 2022-08-24
New Version Script

stevarino:
Put together a quick version updater script for adding a new version to minecraft-data.

My 1.19.21 PR (https://github.com/PrismarineJS/minecraft-data/pull/622) was put together with just this script. Not pretty, but it works. Happy to iterate on style if needed.

Currently untested for pc platform.

### DIFF
--- a/tools/js/incrementVersion.js
+++ b/tools/js/incrementVersion.js
@@ -13,19 +13,19 @@ function writeJSON (path, data) {
   fs.writeFileSync(path, JSON.stringify(data, null, 2), 'utf-8')
 }
 
-function alterJSON(path, callback) {
+function alterJSON (path, callback) {
   const jsonContents = getJSON(path)
   if (callback(jsonContents) !== false) {
     writeJSON(path, jsonContents)
   }
 }
 
-if (process.argv.length != 5) {
+if (process.argv.length !== 5) {
   console.log(`Usage: npm run version <${platforms.join('|')}> {version} {protocol_version}`)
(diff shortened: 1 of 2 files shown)


tools/js/incrementVersion.js
```diff
@@ -0,0 +1,85 @@
+const fs = require('fs')
+const { join } = require('path')
+
+const root = join(__dirname, '..', '..')
+const data = join(root, 'data')
+const platforms = ['bedrock', 'pc']
```
extremeheat, line 6: Only bedrock uses the YAML files right now so this list should only contain bedrock

extremeheat (commented):
Should be added to the updating documentation

tools/js/incrementVersion.js
stevarino: Re: documentation - fully agree. But where should this be documented? minecraft-data, node-minecraft-data, or bedrock-protocol. To me I feel bedrock-protocol makes the most sense as that is where someone would likely look, but could either be standalone or a link to bedrock-specific instructions in minecraft-data. WDYT?

stevarino:
Left a stub in case someone wants to add pc data later, as there are a lot of similar steps in pc version incrementing.

extremeheat:
Lgtm, but have not tested yet.

rom1504:
let's merge after it has been tried at least once

extremeheat (approved):
LGTM, tested script and seems to work

PR #628, opened 2022-09-06
add version.json for 22w13oneblockatatime

Bayugoon:
Some of the protocol might be (probably is) different but this at least allows for a client to connect to the server.


Bayugoon:
> You need to add to the data paths and pc/common

I added it to data path, it seems to already be in pc/common

Bayugoon:
> > You need to add to the data paths and pc/common
> 
> I added it to data path, it seems to already be in pc/common, is there anything else?

it's in one of the files but not version.json, I'll add it

PR #629, opened 2022-09-06
add aprilfools version 22w13oneblockatatime to pc

Bayugoon:


Bayugoon:
block and item data still needs to be updated, I've generated the report files for the version and viewed [burger](https://pokechu22.github.io/Burger/22w13oneblockatatime.json) but I'm not sure how to get all the information needed

Bayugoon:
blocks need to be updated because there is 2210 added blockstates after Air (id: 1 - 2210) for the **generic_item_block** as well as the item form that block that is added after Air (id: 1), so technically it could be possible to just shift the block states of 1.18.2 blocks/items instead of regenerating the files assuming everything else is in the same order.

Bayugoon:
the **item** block state is the protocol id of the item that the ``generic_item_block`` is rendering so the block state can easily be converted to the item resource location/data object

rom1504:
can you please say why we should add this ?

PR #631, opened 2022-09-11
bedrock/1.19.30 initial commit

stevarino:
Still needs work on bedrock-protocol end.

Used https://github.com/Sandertv/gophertunnel/pull/146 as a guide, but skipped add_player and update_abilities packet changes as they seem to be just subtyping and nothing functional.


data/bedrock/latest/proto.yml
```diff
@@ -3726,3 +3732,41 @@ packet_feature_registry:
...
+   client_protocol: i32
+
+packet_game_test_request:
+   !id: 0xc2
+   # MaxTestsPerBatch ...
+   max_tests_per_batch: varint
+   # Repetitions represents the amount of times the test will be run.
+   repetitions: varint
+   # Rotation represents the rotation of the test. It is one of the constants above.
+   rotation: u8
```
extremeheat, line 3754: we can add the constants in https://github.com/Sandertv/gophertunnel/pull/146/files#diff-7da5dae79269c9351484664f4b134a04deafc18b605327b4ee649659907d078cR5 for clarity that this is not an analog field

extremeheat (commented):
missing 
* item_stack https://github.com/Sandertv/gophertunnel/pull/146/files#diff-3d0d084d8e48bea087ffcc600251c0864a51c85f0e8dd3901e4b7452892e5c95

data/bedrock/latest/proto.yml
stevarino: Done and done. Good catch!

Re game_test_request.rotation: 

1) okay... i guess that works. I suppose they would have to step up to 16-bit if they wanted 360 degrees.
2) Kept them as indexed integers. Would strings be better? or perhaps qualified strings (`"rotation0"`)?:
3) 0... and 360?

extremeheat: Yeah, using numbers in the enum isn't good because that can create data type collisions. As users can specify the string enum representation *or* specify a numeric value for these fields, this means ambiguity when serializing in protodef. I've corrected them to be strings.

extremeheat:
Thanks for the work on this!

Issue #636, opened 2022-09-30
Get 1.19 recipes


rom1504:
paths:
* burger
* parse mc jar recipes

rom1504:
task: run https://github.com/Pokechu22/Burger  ; check if output recipes is good

rom1504:
then use https://github.com/PrismarineJS/burger-extractor on the output

PR #638, opened 2022-09-30
added Burger extracted recipes.json for 1.19

FCKJohni:
Title says it all :D

One step closer to get Mineflayer going for 1.19


FCKJohni:
closes #636 when merged

PR #649, opened 2022-10-30
Update for 1.19.2

jtsiskin:
image

test in mineflayer `npm run mocha_test -- -g "mineflayer_external 1.19.2"`

PR #650, opened 2022-10-30
Update to 1.19.2

mygizli04:
I am aware that #649 is open, but I also spent some time on this and would like to have it checked out as well. From my testing the filter type mask, network name and network target name seem to not exist as well.

Credit to @mat-1 for telling me about the player chat message and chat header packets which he let me use from https://github.com/mat-1/azalea


mygizli04:
I'm not sure what the problem with the CI is, I didn't touch items or blocks. I just copied them over to the 1.19.2 folder.

rom1504:
Could you make one commit where you only copy 1.19 protocol to 1.19.2 file without change and another commit with change?
It makes it possible to review the PR

PR #652, opened 2022-11-02
bedrock spawn_particle_effect: add missing bool field right before `molang_variables`

famima65536:


extremeheat:
Please update the protocol JSON files with npm run build in tools/js folder.

rom1504:
old and failing

PR #656, opened 2022-11-29
Test: Find versions to test iteratively

u9g:


rom1504:
no let's instead tests the hardcoded list is complete

PR #657, opened 2022-12-01
initial bedrock 1.19.50 support

WillQizza:
Initial support for 1.19.50
I used both gophertunnel and Cloudburst's protocol for guidance on protocol changes.
However, it is lacking the entity metadata/flag changes.

If necessary, I am willing to implement them, however I do need guidance on how to go about differentiating between MetadataFlags1and MetadataFlags2 as Cloudburst does not seem to differentiate between the two.


JustTalDevelops:
> If necessary, I am willing to implement them, however I do need guidance on how to go about differentiating between MetadataFlags1and MetadataFlags2 as Cloudburst does not seem to differentiate between the two.

Most of the entity metadata changes within Gophertunnel were related to code cleanup in general, nothing specific to the protocol update. The enums themselves did get shifted, but otherwise they're the same.

extremeheat:
Thanks for the work on this ! I've updated the remaining enums.

extremeheat (approved):

PR #662, opened 2022-12-19
Added a new wrapper for C#

psu-de:


rom1504:
can you move that to Projects using minecraft-data section ?

looks like it's more than a wrapper

PR #671, opened 2023-01-19
Rename 1.19.2 chat_command field

extremeheat:
Be clear it's last rejected message not last received message


rom1504:
can you do 1.19.3 too ?

frej4189:
> can you do 1.19.3 too ?

This field is not in 1.19.3

PR #684, opened 2023-02-19
Update bedrock packet_command_request.version comment

stevarino:
Determined via relay. Commands are not executed when set to 0, 1, or 51.

This may be a new field that needs to be added to `common/protocolVersions.json` but its probably too soon to say for sure.

### DIFF
--- a/data/bedrock/latest/proto.yml
+++ b/data/bedrock/latest/proto.yml
@@ -1780,7 +1780,8 @@ packet_command_request:
    # Internal specifies if the command request internal. Setting it to false seems to work and the usage of
    # this field is not known.
    internal: bool
-   # Version is the version of the command that is being executed. This field currently has no purpose or functionality.
+   # Version is the version of the command that is being executed. This field currently has no known purpose
+   # or functionality but should be set to 52 as of 1.19.62.
    version: varint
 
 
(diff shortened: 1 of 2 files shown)


data/bedrock/latest/proto.yml
```diff
@@ -1780,7 +1780,8 @@ packet_command_request:
    # Internal specifies if the command request internal. Setting it to false seems to work and the usage of
    # this field is not known.
    internal: bool
-   # Version is the version of the command that is being executed. This field currently has no purpose or functionality.
+   # Version is the version of the command that is being executed. This field currently has no known purpose
```
extremeheat, line 1783: The comment here is a bit contradictory. The version is important and should be set to a specific value for the version

extremeheat: Something like "Specifies the version of the command to run, relative to the current Minecraft version. Should be set to 52 as of 1.19.62" I think would be clearer

stevarino: Sounds good to me. Hopefully this doesn't change per version though.

extremeheat (approved):

extremeheat:
Thanks!

PR #685, opened 2023-02-19
1.19.3 protocol updates

frej4189:


data/pc/1.19.3/protocol.json
```diff
@@ -3707,7 +3707,7 @@
             },
             {
               "name": "flags",
-              "type": "i8"
+              "type": "u8"
```
extremeheat, line 3710: Why the data type changes? Does it fix anything broken? If it's applicable to the previous versions you'd have to go back and change all the previous versions' types also.

frej4189: I went through the entire protocol packet by packet and just changed the types as I came across them. As for the ubyte/byte that probably won't make any difference as they usually only require the LSBs, but I figured it makes more sense to use the same types as the Vanilla client.

rom1504: Do previous versions need the same change?

frej4189: To match Vanilla implementation, yes.

rom1504: then can you please change it all ? (or revert for this version)

rom1504:
Let's not

frej4189:
This PR also fixes other things. Closing it because of a semantic disagreement doesn't really make sense

rom1504:
Ah yes I didn't see the sound event part

Could you revert the i8/u8 part ? (Or apply it in other versions)

TheRedXD:
I fixed the soundEvent issue in 1.19.3, 1.19.4 and 1.20
The data type change is also not there, for now at least
This is the pull request that changes it: https://github.com/PrismarineJS/minecraft-data/pull/731

rom1504:
I merged the other PR

PR #682, opened 2023-02-17
Add bedrock 1.19.62 protocol data

CreeperG16:
Bedrock 1.19.62

PR #686, opened 2023-02-24
Add Bedrock/1.19.63 (Protocol 568)

stevarino:
Looks like just a version bump to protocol 568.


extremeheat:
If it's just a version bump (no changes to protocol) we don't need new protocol files, adding to the dataPath and protocolVersions.json file should work. The bump was expected in #682

data/bedrock/latest/proto.yml
```diff
@@ -1,7 +1,7 @@
 # Created from MiNET and gophertunnel docs
 # The version below is the latest version this protocol schema was updated for.
 # The output protocol.json will be in the folder for the version
-!version: 1.19.63
+!version: 1.19.62
```
stevarino, line 4: Just curious, what's the logic about reverting this field? Is it used anywhere or just git log cleanliness?

extremeheat: It's used when determining where the protocol.json file will be written (which is ultimately used by protodef)

extremeheat (approved):

extremeheat:
Thanks!

PR #691, opened 2023-03-02
Update worldparticles packet for 1.19.3

u9g:
For what the changes should be, see: https://wiki.vg/Protocol#Particle

### DIFF
--- a/data/pc/1.19.3/protocol.json
+++ b/data/pc/1.19.3/protocol.json
@@ -167,7 +167,7 @@
               }
             ]
           ],
-          "24": [
+          "25": [
             "container",
             [
               {
@@ -176,7 +176,7 @@
               }
             ]
           ],
-          "35": [
+          "39": [
             "container",
             [
               {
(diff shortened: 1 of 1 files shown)


frej4189:
Should probably be backed up by source code aswell. Wiki is not up to date for 1.19.3

extremeheat:
I checked the data generator reports, and seems to be correct. But this should also be applied to 1.19 and 1.19.2. 

Particle types for 1.19.2 from data generator report :  https://gist.github.com/extremeheat/9609186e0087a228c0cf1627ee578bff

extremeheat (approved):

u9g:
to my future self, heres how to apply a diff on one file to other versions:
1. Get the .diff from `wget https://github.com/PrismarineJS/minecraft-data/commit/7db6bef15980ec9a0cc670bf150575ec277376e1.diff` but with a different commit
2. For the path of the file changed, use `git diff-tree --no-commit-id --name-only "$(git rev-parse HEAD)" -r`
3. Apply patch with `patch -p1 data/pc/1.19/protocol.json ./7db6bef15980ec9a0cc670bf150575ec277376e1.diff` but with a different commit and file

PR #693, opened 2023-03-03
Change can_place_on and can_destroy to arrays of strings

CreeperG16:


extremeheat:
Is this correct and working ? Does not seem like it should be

CreeperG16:
sorry I did that wrong

extremeheat:
Done in #694

PR #709, opened 2023-06-03
node 18

rom1504:


rom1504:
idk where is that require-self mention

extremeheat:
it's from node-minecraft-data https://github.com/PrismarineJS/node-minecraft-data/blob/master/package.json#LL13C17-L13C80

None of the prepare scripts seem to need it so it can probably be removed, but the problem maybe it's not installing the devDeps for some reason

rom1504:
we can simply replace require self by self dep
https://github.com/PrismarineJS/node-minecraft-data/pull/286

rom1504:
I don't understand why it fails with `npm ERR! Error: Cannot find module 'json-schema-to-typescript'`

rom1504:
so error here is still `npm ERR! Error: Cannot find module 'json-schema-to-typescript'`

tried various things but nothing helps

any more idea here @extremeheat ?

extremeheat:
Should be fixed, not sure what flow was going on previously

rom1504:
Thanks

PR #710, opened 2023-06-05
pc: add entity metadata

extremeheat:
Adds entity metadata keys to entities.json for pc, updates protocol.json to use string mapping for metadata item values

Addresses #72 a bit

### DIFF
--- a/tools/js/extractPcEntityMetadata.js
+++ b/tools/js/extractPcEntityMetadata.js
@@ -22,7 +22,6 @@ function getEntityTypes () {
       const regex = line.match(/EntityType<(.*)> (.*) = register\(\W+"([a-z0-9_]+)"/)
       if (regex) {
         const [, type, , name] = regex
-        // console.log('Type', type, 'Name', name)
         classNameTo[type] = name
       }
     }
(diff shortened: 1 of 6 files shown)


tools/js/extractPcEntityMetadata.js
```diff
@@ -0,0 +1,153 @@
+const fs = require('fs')
```
rom1504, line 1: I'd rather we have a dedicated package for extraction. tools/js should not contain large amount of code as it's not documented and generally not maintained.

where would that kind of extraction fit? what's the source of data ?

extremeheat: Yeah, I didn't know where else to put it so I left it in here for now.

It basically clones a copy of the mc source code and then does some string manipulation to extract the relevant information from the human readable source code files. Then updates the files in mc-data.

http://github.com/prismarinejs/mcdataextracting requires a fabric mode and Java code injection
http://github.com/prismarinejs/minecraft-jar-extractor is a better fit, but would need some refactoring to use mojang mappings and the [ForgeFlower](https://github.com/MinecraftForge/ForgeFlower) decompiler

rom1504: yeah I was not saying reimplement it another way, just move it elsewhere

maybe it could just be put in minecraft-jar-extractor with no change

rom1504: but let's not block on that, can be done in another pr

PR #716, opened 2023-06-10
Fix pages build failing due to bundle size

extremeheat:
Shrinks the uncompressed bundle size ~2x by removing big `['steve.json', 'blockMappings.json', 'blockCollisionShapes.json', 'blocksJ2B.json', 'blocksB2J.json']` files which are not currently used in the pages build via a browserify transform. In a production web server with compression the bundle size is now 8MB

`-rw-r--r--   1 gitpod gitpod 131M Jun 10 02:20 bundle.js` old with all JSON data
`-rw-r--r--   1 gitpod gitpod  75M Jun 10 02:41 bundle3.js` new stripped
`-rw-r--r--   1 gitpod gitpod  47M Jun 10 02:45 bundle3.min.js` new minified, safely storable in repo (<100mb)
`-rw-r--r--   1 gitpod gitpod   8M Jun 10 02:46 bundle3.zip` new with gzip

Also fixes an issue with pc 1.19.4 biomes causing table gen errors and some bedrock protocol links being 404

### DIFF
--- a/buildBundle.js
+++ b/buildBundle.js
@@ -1,0 +1,25 @@
+const fs = require('fs')
+const browserify = require('browserify')
+const through = require('through')
+
+const blockedFiles = ['steve.json', 'blockMappings.json', 'blockCollisionShapes.json', 'blocksJ2B.json', 'blocksB2J.json']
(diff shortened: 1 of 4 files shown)


rom1504:
thanks

PR #725, opened 2023-06-30
Update bedrock subchunk_request.requests and packet_subchunk.entries for 1.18.11+

irkmandeer:
The type for `dx`, `dy`, `dz` for the client `subchunk_request` and server `subchunk` packets should be a signed byte. 

`u8` was changed to `i8` in `proto.yml`, compiled, and tested against vanilla bedrock servers for all supported versions.


extremeheat (approved):

extremeheat (commented):
Thanks !

PR #726, opened 2023-07-02
Fixed typo in dataPaths.json for bedrock 1.19.21

irkmandeer:
Fixed a typo that causes `1.19.21` to use the `protocol.json` of `1.19.20`.

This is somewhat pointless since the compiled `protocol.json` files for `1.19.20` and `1.19.21` are identical


extremeheat:
So yeah since there was no changes we could probably not have a protocol.json for that version at all, and just direct to the last versions data. I forgot to comment that on the PR for that update. But since we have anyway we should probably point to it.

extremeheat (approved):

PR #727, opened 2023-07-02
Add commands workflow

extremeheat:


.github/workflows/commands.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Check out repository
+      uses: actions/checkout@v3
+    - name: Run command handlers
+      uses: PrismarineJS/prismarine-repo-actions@master
+      with:
+        # NOTE: You must specify a Personal Access Token (PAT) with repo access here. While you can use the default GITHUB_TOKEN, actions taken with it will not trigger other actions, so if you have a CI workflow, commits created by this action will not trigger it.
+        token: ${{ secrets.PAT_PASSWORD }}
+        # See `Options` section below for more info on these options
+        install-command: npm install
+        /fixlint.fix-command: npm run fix
```
rom1504, line 22: are you sure this will work here?

mcdata is a bit different from all other repos

there's tools/js for package.json stuff

and for releases there's only HISTORY.md , no package.json

extremeheat: Yeah, the package.json is optional, it can read the version from just the history https://github.com/extremeheat/minecraft-data/pull/17

rom1504: how does it know where to run npm run fix ?

extremeheat: It runs the commands specified in the workflow options at the root of the repo. But yeah it was missing the “cd tools/js” step, fixed that in the workflow options

rom1504:
nice, seems to work

maybe let's add it in prismarine template
and let's use https://github.com/PrismarineJS/prismarine-meta to open PRs to add it everywhere

rom1504:
btw I created https://github.com/PrismarineJS/minecraft-data/issues/728 to issue the commands

does that make sense? should we do this everywhere?

extremeheat:
Yeah, it can be added probably to most of the repos. It's not required to make a new issue, it should work in any issue or PR comment if the user is the author or has commit permissions on the repo.

PR #730, opened 2023-07-03
Bedrock protocol changes for 1.17.0 to 1.20.0

irkmandeer:
The following changes affects all versions of bedrock, and is needed for breaking blocks, interacting with containers, etc. when server authoritative movement is enabled.

### TransactionUseItem

The `block_position.y` value in`packet_player_auth_input.transaction.data` needs to be a `varint`, and `packet_player_auth_input.transaction.data.face` needs to be `zigzag32` 
- The type for `block_position` was changed from `vec3i` to `BlockCoordinates`  in `types.yml`
- The type for `face` was changed from `varint` to `zigzag32` in `types.yml`

### Action

The `position.y` value in `packet_player_auth_input.block_action[]` needs to be `zigzag32` for all actions
- The type for `position` was changed from `BlockCoordinates` to `vec3i` in `proto.yml`

The updates was tested for for all supported 1.17.0+ versions against vanilla bedrock servers.

The 1.16.x versions was skipped due to issues that needs to be resolved first:

- 1.16.201 - Broken block palette, protocol runtime errors. Relay gets kicked due to `packet_violation_warning`
- 1.16.210 - Broken block palette, protocol runtime errors. 
- 1.16.220 - This version actually works, but guess there is no point in including just this one.

### DIFF
--- a/data/bedrock/1.16.210/proto.yml
+++ b/data/bedrock/1.16.210/proto.yml
@@ -2271,7 +2271,7 @@ packet_player_auth_input:
             if start_break or abort_break or crack_break or predict_break or continue_break:
                # BlockPosition is the position of the target block, if the action with the ActionType set concerned a
                # block. If that is not the case, the block position will be zero.
-               position: BlockCoordinates
+               position: vec3i
                # BlockFace is the face of the target block that was touched. If the action with the ActionType set
                # concerned a block. If not, the face is always 0.
                face: zigzag32
--- a/data/bedrock/1.16.210/protocol.json
+++ b/data/bedrock/1.16.210/protocol.json
@@ -772,7 +772,7 @@
         },
         {
           "name": "face",
-          "type": "varint"
+          "type": "zigzag32"
         },
         {
           "name": "hotbar_slot",
@@ -7229,7 +7229,7 @@
                                   [
                                     {
                                       "name": "position",
-                                      "type": "BlockCoordinates"
+                                      "type": "vec3i"
                                     },
                                     {
                                       "name": "face",
@@ -7242,7 +7242,7 @@
                                   [
                                     {
                                       "name": "position",
-                                      "type": "BlockCoordinates"
+                                      "type": "vec3i"
                                     },
                                     {
                                       "name": "face",
@@ -7255,7 +7255,7 @@
                                   [
                                     {
                                       "name": "position",
-                                      "type": "BlockCoordinates"
+                                      "type": "vec3i"
                                     },
                                     {
                                       "name": "face",
@@ -7268,7 +7268,7 @@
                                   [
                                     {
                                       "name": "position",
-                                      "type": "BlockCoordinates"
+                                      "type": "vec3i"
                                     },
                                     {
                                       "name": "face",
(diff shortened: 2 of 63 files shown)


extremeheat:
Thanks for the investigation into this. In the future, please link to the source of the changes, as that makes it easier to review the PR.

> 1.16.201 - Broken block palette, protocol runtime errors. Relay gets kicked due to packet_violation_warning
> 1.16.210 - Broken block palette, protocol runtime errors.

Which changes that you introduced cause these problems?

irkmandeer:
I simply ran across the problems while implementing bedrock support for mineflayer (to get myself clued up on the project) .

I've found that varint and zigzag32 problems is the first thing you need to look for. 

I'm working on a simple test framework that test gameplay functionality against multiple versions of bedrock. The workflow for this PR was to write a test after finding the problem, see which versions fails, implement fix, test that all versions are working. 

For 1.16.x, no changes were made. Maybe no-one attempted to test gameplay against them? There is no blocks and blockstates defined for 1.16.201 and 1.16.210. I'll investing a bit more and send details if you want? 

Would it be better to PR a single version in the future, get that reviewed, then do a bulk PR afterwards?

extremeheat:
> For 1.16.x, no changes were made. Maybe no-one attempted to test gameplay against them? There is no blocks and blockstates defined for 1.16.201 and 1.16.210. I'll investing a bit more and send details if you want?

I'm confused here. Are you saying the changes are not applicable to the old versions, or you are just unable to test them? If it's the latter the changes should be applied to all versions and the CI will decide if it's working or not. 

> Would it be better to PR a single version in the future, get that reviewed, then do a bulk PR afterwards?

No, if you make any protocol changes they should be applied to all the versions. It's hard to track otherwise.

irkmandeer:
I was unable to test and verify 1.16.101 and 1.16.210 due to errors, and being kicked by the server:
- 127.0.0.1/50023 Read failure for 0x3f <Buffer 3f 01 01 58 39 a4 55 0a 15 85 48 8d e9 08 8f fb dd 7f b9>  127.0.0.1/50023 PartialReadError: Read error for undefined : undefined
- Kicked  127.0.0.1/50023 **Server packet parse error**
- [serverbound] **packet_violation_warning** Connection closed:  127.0.0.1/58834 undefined

However, with some patience, I was able to get a working block palette, and was able to update `1.16.210` and  `1.16.220`. I pushed the changes into the branch for this PR.

From what I can tell no change is needed for `1.16.201`. The vanilla client does not appear to be sending `player_auth_input` containing transactions and block_actions. There are also no `item_interact` and `block_action` input flags defined in the protocol for `packet_player_auth_input.input_data`

extremeheat:
OK, the PR looks good to me. The protocol data for 1.16 may be a bit incorrect indeed, as it's an early supported version and we didn't have much testing infra at the time to ensure the types were all correct; instead the protocol got refined for correctness over time. The best way to fix would be to backward diff the versions and backport any applicable changes, but this is a separate topic to this PR.

extremeheat (approved):

extremeheat:
Thanks for the contribution!

PR #738, opened 2023-07-18
damage_event packet fix

PondWader:
As can be seen here https://wiki.vg/Protocol#Damage_Event, `sourceCauseId` and `sourceDirectId` are not optional, they are always sent, just if not present set to 0.

This applies to both 1.20 and 1.19.4 when the packet was introduced. It caused node-minecraft-protocol to display a PartialReadError whenever the client was hit

### DIFF
--- a/data/pc/1.19.4/protocol.json
+++ b/data/pc/1.19.4/protocol.json
@@ -5378,17 +5378,11 @@
             },
             {
               "name": "sourceCauseId",
-              "type": [
-                "option",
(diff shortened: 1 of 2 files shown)


frej4189:
Verified with source.

https://github.com/extremeheat/extracted_minecraft_data/blob/client1.19.4/client/net/minecraft/network/protocol/game/ClientboundDamageEventPacket.java

PR #734, opened 2023-07-12
Bedrock 1.20.10 protocol support

CreeperG16:
- Bumped verison to 1.20.10, protocol version to 594
- Changed `charged_crossbow` event ID to `charged_item`
- Updated `available_commands` packet to support `chained_subcommands`
- Ddded `missed_swing`, `start_crawling` and `stop_crawling` input flags and player actions
- added `agent_animation` packet (0x130)
- added `collision_box` metadata key, `crawling` metadata flag

PR #742, opened 2023-07-22
Fix values for enum type in available_commands packet

CreeperG16:
The bedrock `available_commands` packet's overloads field has a value called `enum_type`, which is documented to be set to possible values 0x10, 0x20, 0x100 and 0x400; however upon testing all supported versions on their vanilla BDS, two of these values seem to be incorrect. 0x20 should be 0x30 (48) and 0x400 should be 0x410 (1040). This PR is a fix for these values.
There also seemed to be something strange with the `overloads` field in `available_commands` - In 1.20.10, the objects in the array have the `chaining` bool, but then the parameters are in another field called `__4265`. I assume this was because it was setting an array as an anonymous field. I fixed this by naming the field `parameters` - but if this breaks anything I can revert it. Locally testing it with a vanilla BDS on a few versions including 1.20.10 looks good.

### DIFF
--- a/data/bedrock/1.16.201/proto.yml
+++ b/data/bedrock/1.16.201/proto.yml
@@ -1332,7 +1332,7 @@ packet_available_commands:
       # The list of overload parameters for this command
       overloads: []varint
          # Each of the parameters gets an array of posible overloads
-         _: []varint
+         parameters: []varint
             # The name of the parameter shown to the user (the `amount` in `/xp <amount: int>`)
             parameter_name: string
             value_type: lu16 =>
--- a/data/bedrock/1.16.201/proto.yml
+++ b/data/bedrock/1.16.201/proto.yml
@@ -1352,9 +1352,9 @@ packet_available_commands:
             # In MC, this + prior field are combined to one 32bit bitfield
             enum_type: lu16 =>
                0x10: valid
-               0x20: enum
+               0x30: enum
                0x100: suffixed
-               0x400: soft_enum
+               0x410: soft_enum
             # Is this parameter required?
             optional: bool
             # Additinal options for this command (thanks macroshaft...)
--- a/data/bedrock/1.16.201/protocol.json
+++ b/data/bedrock/1.16.201/protocol.json
@@ -5376,66 +5376,74 @@
                       {
                         "countType": "varint",
                         "type": [
-                          "array",
-                          {
-                            "countType": "varint",
-                            "type": [
-                              "container",
-                              [
+                          "container",
+                          [
+                            {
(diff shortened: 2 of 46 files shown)


extremeheat:
> then the parameters are in another field called __4265

Where do you see that?

CreeperG16:
I started a vanilla BDS on version 1.20.10, connected a bedrock-protocol client, and just looked at `packet.command_data[i].overloads`

CreeperG16:
I started a vanilla BDS on version 1.20.10, connected a bedrock-protocol client, and just looked at `packet.command_data[i].overloads`

extremeheat:
Can you provide some deserialized output with that number field?

CreeperG16:
Sorry for the delay
Connecting a client to a vanilla 1.20.10 BDS, here is the first entry in `packet.command_data`:
```js
> console.log(until.inspect(packet.command_data[0], { depth: 4 }))
{
  name: 'allowlist',
  description: 'commands.allowlist.description',
  flags: 0,
  permission_level: 4,
  alias: 182,
  chained_subcommand_offsets: [],
  overloads: [
    {
      chaining: false,
      __4265: [
        {
          parameter_name: 'action',
          value_type: 181,
          enum_type: 48,
          optional: false,
          options: [Object]
        },
        {
           parameter_name: 'name',
           value_type: 'string',
           enum_type: 'valid',
           optional: true,
           options: [Object]
        }
      ]
    }
  ]
}
```
In versions before 1.20.10 (when the `chaining` bool field was added), the entry looks like this:
```js
{
  name: 'allowlist',
  description: 'commands.allowlist.description',
  flags: 0,
  permission_level: 4,
  alias: 182,
  chained_subcommand_offsets: [],
  overloads: [
    [
      {
        parameter_name: 'action',
        value_type: 181,
        enum_type: 48,
        optional: false,
        options: [Object]
      },
      {
         parameter_name: 'name',
         value_type: 'string',
         enum_type: 'valid',
         optional: true,
         options: [Object]
      }
    ]
  ]
}
```

extremeheat:
Ah, I see. This problem was introduced with 1.20.10 PR #734, it's caused by a `overloads` no longer being possible to be an anonymous array with the introduction of a [new field `overload` field](https://github.com/PrismarineJS/minecraft-data/commit/fa521866fb2c50af30d072413b2a12a5a2a5eb45#diff-9c62d46fc27e8b994cb346d654d61abca063191abf9762890a12659b56048966L1735-R1759) preceding the nested array (basically we can't just flatten it anymore). So I think it makes sense to remove the anon there (and on all the other versions too to be consistent).

PR #746, opened 2023-07-30
fix: correct schema types

Eejit43:
This PR makes two fixes:

* Adds types to `stackSizeRange` in the blockLoot schema, setting the `number` or `null` type.
* Marks `id` and `name` as required in the particles schema, and sets `additionalProperties` to false.


Eejit43:
Ah it appears it can be nullable. I'll fix that shortly.

PR #749, opened 2023-08-08
Bedrock stack network id protocol fix for 1.16.220 to 1.20.10

irkmandeer:
The `stack_id` field in `StackRequestSlotInfo` and `Item` needs to be `varint` and not `zigzag32`.

This causes a mismatch of the `stack_id` for an item stack between the `inventory_content`, `inventory_slot`, `inventory_transaction`, `item_stack_request` and `item_stack_response` packets.

The problem can be seen by using the relay to print the `inventory_content`, `item_stack_request` and `item_stack_response` packets; moving an item in the inventory, then comparing the `stack_id` from the packets

Code references 
- [CloudburstMC](https://github.com/CloudburstMC/Protocol/blob/e2abbaed93d8e19f4db62c51567dea7c3496c9ed/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/codec/v422/serializer/ItemStackResponseSerializer_v422.java#L19)
- [PMMP](https://github.com/pmmp/BedrockProtocol/blob/master/src/types/inventory/stackresponse/ItemStackResponseSlotInfo.php#L45)

The change was applied and tested for for all supported (and playable 1.16.220+) versions against vanilla bedrock servers.


extremeheat (approved):

extremeheat:
Actually, I am confused here. Both your links do not show an unsigned varint, they show a signed varint, which is what the current field is. How are you sure this field is not wrong as opposed to the others?

irkmandeer:
You are absolutely correct! 

I specifically check the other implementations since both types work, just as long as they are consistent. I've got no idea why my brain decided on varint when it's clearly zigzag and network ids always seems to be zigzag. 

I wrote tests for all this, so let me change it to zigzag, and run everything again.

Do I need to close this PR and open another one?

extremeheat:
> Do I need to close this PR and open another one?

Up to you

irkmandeer:
Sorry for the wait, I was out of action the last couple of week.

I just created a new PR [762](https://github.com/PrismarineJS/minecraft-data/pull/762) for this.

PR #763, opened 2023-08-25
add features for prismarine chunk

extremeheat:
add some features from prismarine-chunk

### DIFF
--- a/data/bedrock/common/features.json
+++ b/data/bedrock/common/features.json
@@ -1,5 +1,15 @@
 [
   {
+    "name": "smallWorld",
+    "description": "world height is 128 blocks",
+    "version": "0.14.3"
+  },
+  {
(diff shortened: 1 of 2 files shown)


extremeheat:
Could be helpful to replace "versions" field with "versionRange" and add a "versions" with only explicit versions

rom1504:
Ah do you have some case where the version list does not form a single range ?

extremeheat:
I thought pchunk needed it but it seems not, but prismarine-physics does have split ranges in its features.json

rom1504:
I see, seems quite surprising

rom1504:
but yeah we could bring the feature maybe
not 100% sure if it helps though as it seems it's an edge case so maybe it'd be more confusing than helping

PR #767, opened 2023-08-27
Handle snapshots in update checking workflow

extremeheat:


rom1504:
how do you test this?

extremeheat:
It can be run locally to test the basic functionality, and it can be tested on github by switching the default branch on a fork and manually triggering with a workflow dispatch like in https://github.com/extremeheat/minecraft-data/actions/runs/5991545064/job/16249992745

rom1504:
I think it would be great to add a readme in that folder explaining how to run it

eventually we can move this to its own repo

PR #769, opened 2023-09-05
fix bedrock 0.14/0.15 data paths

extremeheat:
using major version (0.14/0.15) instead of mc version causes version comparisons to break


extremeheat:
Actually, it seems like alot of versions also have this problem. Likely would be better to handle in node-mc-data

PR #770, opened 2023-09-09
Change has_stack_id in bedrock Item to a boolean

CreeperG16:
Change the type of the `has_stack_id` field in `Item` type for bedrock protocol definition to a boolean type.

### DIFF
--- a/data/bedrock/1.16.220/protocol.json
+++ b/data/bedrock/1.16.220/protocol.json
@@ -481,7 +481,7 @@
                   },
                   {
                     "name": "has_stack_id",
-                    "type": "u8"
+                    "type": "bool"
                   },
                   {
                     "name": "stack_id",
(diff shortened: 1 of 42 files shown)


extremeheat:
How do you know this is correct?

CreeperG16:
I double checked locally that true corresponds to a 1 and false corresponds to a 0, other than that it just makes more sense. The field *is* named `has_stack_id` after all. Though if there is a reason why it would be a number, my bad that I didn't find it.

extremeheat (approved):
Ah, you seem to be correct. I was mistakenly thinking about at the has NBT flag, which is not a boolean.

AnyBananaGAME:
> I double checked locally that true corresponds to a (byte) 1 and false corresponds to a (byte) 0, other than that it just makes more sense. The field *is* named `has_stack_id` after all. Though if there is a reason why it would be a number, my bad that I didn't find it. Though I'd be curious as to what that reason is.

Yea it is

AnyBananaGAME:
I get that with a relay

extremeheat:
After a lot of investigation, this subtle change seems to be breaking Item serialization. I'm not sure why this wasn't caught by the tests, but this breaks the switch check below as it's doing integer comparisons on the boolean. It's possible the switch below can be fixed, but in the meantime I'll revert this PR.

PR #778, opened 2023-09-21
Add bedrock 1.20.15 protocol data

irkmandeer:
Bedrock 1.20.15 uses the same v594 codec as 1.20.10.

The update follows the same pattern as 1.19.63 which only includes a `version.js` and `proto` in `dataPaths.js` pointing to a previous version.


extremeheat (approved):

extremeheat:
It shouldn't be necessary to add to the data paths if nothing at all changed, but all the versions should be in protocolVersions.json so I'll merge the PR. 

As a side note we should probably have an automated way to update this similar to the pc workflow we have now.

PR #784, opened 2023-10-25
Add support for bedrock 1.20.40 protocol

CreeperG16:
This PR adds data for bedrock 1.20.40 protocol.
Changes:
- `reason` field in disconnect packet, and `DisconnectFailReason` enum for that field. (The name confuses me, why is it a disconnect *fail* reason? But that's what it's called in [CloudburstMC](https://github.com/CloudburstMC/Protocol/commit/bb8e4067b0d21826b7e6910dced986c428a8620f#diff-47e3f111c4991f4179eeac3d92f59625c4199b160f035baab808a8b7ffb438a2R3) )
- `received_server_data` player action and player_auth_input input data
- `timer_flag_1`, `_2` and `_3` entity metadata flags
- MapDecoration types

1.20.30:
- Add missing `start_flying` and `stop_flying` player actions

Notes:
- The disconnect fail reasons are given in ALL_CAPS in [CloudburstMC's code](https://github.com/CloudburstMC/Protocol/commit/bb8e4067b0d21826b7e6910dced986c428a8620f#diff-47e3f111c4991f4179eeac3d92f59625c4199b160f035baab808a8b7ffb438a2) , but I changed them to lower snake_case to match the format of most other data. Should they be reverted to ALL_CAPS?
- On a second look at [Gophertunnel's code](https://github.com/Sandertv/gophertunnel/commit/a3633787911dc7bf0e92ab5010b8c6bb8432810c#diff-dac426c41ce45168e105cbae844414ab13a2bcbccc51b175fee6663b33486a2eL43) , the MapDecoration type field doesn't seem to actually be typed by the constants defined above it; or am I reading it wrong? If so, I'll revert the [change](https://github.com/PrismarineJS/minecraft-data/pull/784/files#diff-b82eacd04d406a674e05c33d492d601752ae9b4174c00940be6490f5befb303bR1178-R1202) to the `type` field; however should these values be provided somewhere else? In a comment above, perhaps.


data/bedrock/latest/types.yml
```diff
@@ -2125,4 +2155,97 @@ CameraPresets:
    rotation: Vec2fopts
    ## TODO: make this an enum afer adding proper optional support inside pdefyaml
    audio_listener?: u8
-   player_effects?: bool
\ No newline at end of file
+   player_effects?: bool
+
+DisconnectFailReason: varint =>
```
extremeheat, line 2160: This should be zigzag32

data/bedrock/latest/types.yml
```diff
@@ -1169,7 +1175,31 @@ TrackedObject:
 # MapDecoration is a fixed decoration on a map: Its position or other properties do not change automatically
 # client-side.
 MapDecoration:
-   type: u8
+   type: u8 =>
```
extremeheat, line 1178: It's ok if you want to do these changes, but if done in this PR then we'd need to add to all the older versions as well

extremeheat:
Thanks for the work

PR #786, opened 2023-10-28
Add 1.20.2 data

wgaylord:
This adds the 1.20.2 data

I updated the packets using wiki.vg as a guide. 

I used minecraft-data-generator-server to generate most of the other data, I had to fix the recipes it outputted since it used -1 instead of null. 

I used mc-data-command-generator to generate the commands.json.

I updated the mapIcons.json by just checking the code for the names of each new icon.

And that's about it.

Hopefully everything looks good. We will need to eventually grab a fresh login packet for the loginPacket.json since there was login packet changes, but that sort of requires using minecraft-protocol.


rom1504:
Ah wow this included the protocol, didn't see.

We'll need some much more extensive review for protocol.json

Will look

PR #791, opened 2023-11-16
Fix errors in protocol.json caused by my horrible spelling

wgaylord:
Fixes miss-spelling of configuration that was breaking its packets.

This is just a basic fix for my dumb mistakes, still have to fully verify the whole protocol.


wgaylord:
I am going to just make a final PR once I find all the spelling issues, since this one seems to not be passing the tests.

PR #792, opened 2023-11-16
1.20.2 - Fix spelling issue in Configuration packets and 1 missed change in a play packet.

wgaylord:
While trying to get node-minecraft-protocol to work on 1.20.2 I discovered I have some horrible spelling mistakes in the Configuration packet section, and while fixing that I decided to look over all the packets and found that the Set Experience Level packet field order had changed except it wasn't documented on wiki.vg at the time I originally updated the protocol.json file.


wgaylord:
No idea why  this think made that huge diff. Closing it for now.

PR #803, opened 2023-12-13
Add 1.20.3 and 1.20.4 extract-able data

wgaylord:
This adds all the extract-able data for 1.20.3 and 1.20.4

Interestingly there were changes to the `block.json` between 1.20.3 and 1.20.4.

Also due to how the newer "experimental" features stuff works I have noticed that the block data includes the 1.21 blocks that are exposed by that experimental switch on world creation. 

The protocol for 1.20.3 and 1.20.4 are currently pointing at 1.20.2, this won't work due to changes in chat messages (Now NBT data instead of JSON) and new packets in the configuration and play states. (Relating to resourcepacks)


wgaylord:
We should hold off on merging this till #804 gets figured out since I used the same to generate this data

rom1504:
what tools did you use for generation ?

wgaylord:
I used https://github.com/PrismarineJS/minecraft-data-generator-server 

I also have a version that includes the `mapicon.json` generation as well.

rom1504:
is this ready to merge now ? looks fine to me

rom1504:
ah please add the versions in the readme

wgaylord:
I will add the versions in the readme. Also this currently does not have a `protocol.json` for 1.20.3/1.20.4 yet as noted in the initial PR message.  I will be making another PR for that to full explain the changes that occured.

rom1504:
thanks

rom1504:
we'll need the protocol next indeed

PR #807, opened 2023-12-18
1.20.2 protocol: nbt data can have different root data types

extremeheat:
After https://github.com/PrismarineJS/prismarine-nbt/pull/79/files

### DIFF
--- a/data/pc/1.20.2/protocol.json
+++ b/data/pc/1.20.2/protocol.json
@@ -432,7 +432,7 @@
           ],
           "block_state": "varint",
           "optional_block_state": "optvarint",
-          "compound_tag": "nbt",
+          "compound_tag": "nbtRoot",
           "particle": "particle",
           "villager_data": [
             "container",
--- a/data/pc/1.20.2/protocol.json
+++ b/data/pc/1.20.2/protocol.json
@@ -28,6 +28,18 @@
     "restBuffer": "native",
     "nbt": "native",
     "optionalNbt": "native",
+    "nbtRoot": [
+      "nbt",
+      {
+        "tagType": "anonymousNbt"
+      }
+    ],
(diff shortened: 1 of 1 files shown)


data/pc/1.20.2/protocol.json
```diff
@@ -28,6 +28,18 @@
     "restBuffer": "native",
     "nbt": "native",
     "optionalNbt": "native",
+    "nbtRoot": [
+      "nbt",
+      {
+        "tagType": "anonymousNbt"
```
rom1504, line 34: where is tagType used?

extremeheat: `nbt` is a custom type in node-minecraft-protocol, tagType is passed down to the type serializer code for calling prismarine-nbt parsePacketBuffer

wgaylord:
I think you may have missed a few places where nbt is the type in the protocol to change to the nbtRoot type.  For example in EntityMetaData.  Since all uses of NBT in the protocol now use the anonymous root tag.

data/pc/1.20.2/protocol.json
```diff
@@ -117,7 +129,7 @@
                     },
                     {
                       "name": "nbtData",
-                      "type": "optionalNbt"
+                      "type": "optionalNbtRoot"
```
frej4189, line 132: Item NBT data is still written with a named root tag.

frej4189 (changes requested):

data/pc/1.20.2/protocol.json
frej4189: Seems I was actually incorrect here. Just very weird naming in Vanilla codebase that had me confused (i.e. `readUnnamedTag` actually reads a named tag, whereas `readAnyTag` reads an unnamed tag)

PR #810, opened 2023-12-26
1.20.2: use new anonOptionalNbt and anonymousNbt types from prismarine-nbt, add loginPacket

extremeheat:
after https://github.com/PrismarineJS/prismarine-nbt/pull/81 which needs https://github.com/PrismarineJS/node-minecraft-protocol/pull/1265


extremeheat:
Also added the loginPacket.json with the codec for 1.20.2, dumped from nmp test output. Though technically the dimensionCodec isn't apart of the login packet anymore, I just left it in here for now for compatibility with existing code

PR #812, opened 2023-12-27
1.20.2 Protocol fixes

wgaylord:
All other new and changed packets now match up to MC source for 1.20.2, I will been checking all the packets in my 1.20.3 PR next. 

Seems also chunk pos what changed, will be fixing that.

### DIFF
--- a/data/pc/1.19.3/protocol.json
+++ b/data/pc/1.19.3/protocol.json
@@ -286,6 +286,24 @@
         }
       ]
     ],
+    "soundSource": [
+      "mapper",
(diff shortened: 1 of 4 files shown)


wgaylord:
Think this is everything, while working on this noticed that 1.19.4 has its packet_chunk_biomes packet incorrect. But thats a problem for another day. 


I feel like we are bound to find other issues cropping up again even after this. Since 1.20.2 was a learning experience for me for updating minecraft-data's protocol files.

PR #829, opened 2024-01-09
Propose blocks extended data

zardoy:
I was trying to extract all interaction shapes I needed with a quick parse script on js and also decided to do the same for other blocks data from 1.20.2 code. I'll try to rewrite setup with burger for reliability & automatization for all versions, but the data will include:
- name
- noOcclusion=true on some blocks (essentials for p viewer)
- lightLevel (actually can already be extracted with burger) (essentials for p viewer)
- isAir=true
- isLiquid=true
- isReplacable=true (essential for flying squid)
- color - RGB color of block (I want it most, to be honestly)
- instrument for noteblock

These additional data may be large, should they be merged with blocks.json? Is this is good direction overall?

PR #832, opened 2024-01-10
Propose interaction shapes data

zardoy:
similar to https://github.com/PrismarineJS/minecraft-data/pull/829 want to propose having interaction shapes which is essential for p web client and bot.cursorAtBlock to work correctly. As far as I remember its currently not possible to detect/break blocks like redstone wires with mineflayer.
I extracted interaction shapes of all interested blocks (with a few per-class manual fixes). It was much harder to extract using text manipulation since all blocks set these shapes differently and some like redstone calculate them at class initialization, I'm pretty sure it would be much easier to extract using them using other extractors, but I can't investigate it, sorry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Data provided:
 | Foods | list of foods each with there id, saturation, foodpoints and more |
 | Commands | a tree structure for vanilla minecraft server commands, and some info needed to implement sub-parsers.
 | Legacy | mappings between legacy (1.12) and post-flattening (1.13+) blocks and items ids
+| More interaction shapes | Extended shapes in addition to the collision shapes, for example the torch or redstone cursor shapes.
 | Skin data | (bedrock edition) Skin geometry and texture data for steve skin
 | Features | This can be used to check is a specific feature is available in the current Minecraft version. This is usually only required for handling version-specific functionality.
 
--- a/data/pc/common/moreInteractionShapes.json
+++ b/data/pc/common/moreInteractionShapes.json
@@ -1,0 +1,6593 @@
+{
+    "moving_piston": [],
+    "water": [],
+    "lava": [],
+    "bubble_column": [],
+    "air": [],
+    "void_air": [],
+    "cave_air": [],
+    "mangrove_propagule": {
+        "hanging=true": [
+            7,
+            0,
+            7,
+            9,
+            16,
+            9
+        ],
+        "age=0": [
+            7,
+            13,
+            7,
+            9,
+            16,
+            9
+        ],
+        "age=1": [
+            7,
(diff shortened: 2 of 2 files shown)


zardoy:
About the format. Item can be either:

- single shape (0-16 values)
- object with states e.g. with keys like `north=side|up,...`. use "" state if nothing matches
- object with `"combine"` key e.g. redstone_wire or fire uses it to combine a few shapes

[parser](https://github.com/zardoy/prismarine-block/blob/00cd810ca6853024b2e73ff0d405d1b1e397defc/index.js#L196-L221)

The only known block that needs special shape calculation at runtime is `light` block.

rom1504:
1. this should be per version ; this is not data that is cross version
2. what is missing from https://github.com/PrismarineJS/minecraft-data/blob/48589a6ad3276556a7ea38f219e079c4a7c9017e/data/pc/1.20/blockCollisionShapes.json and can the format be similar ?

zardoy:
> 1. this should be per version ; this is not data that is cross version

yes, that's what I'm saying, I don't have a way to automatically extract it from any version. That included manual work, that's why I can propose data for the latest version only. I'm personally using it with any version with the block renames data without any problems so far.

btw Im just curious do any of these block shapes change over versions?

> 2\. what is missing from [`48589a6`/data/pc/1.20/blockCollisionShapes.json](https://github.com/PrismarineJS/minecraft-data/blob/48589a6ad3276556a7ea38f219e079c4a7c9017e/data/pc/1.20/blockCollisionShapes.json) and can the format be similar ?

this file is missing redstone wire for example since this includes collision shapes only, not cursor interaction. I'm pretty sure we can make the format compatible here if make it per version. On the other hand the collision block data format would be much easier to understand if it used the same per-state object format that is used in blockStates.json in mc assets e.g. key with properties - data (instead of an array of variation). Anyway, I can't make it compatible if this is the only requirement for now because of this.

I'm not doing the tool to make this file, that's what I have for now..

@wgaylord Also I believe we have discussed this as well, but I don't remember what you said about per-version extraction

Gjum:
Hi, creator of `blockCollisionShapes.json` here. That format was made to work with pre-flattening (numeric id:meta) identifiers; if there's a way to use the `block[foo=bar]` format with pre-flattening then that would be preferable for both collision and interaction shapes as it's much more human readable. And yes, collision (physics) and interaction (digging) shapes are different and need separate data files, and also change between versions (e.g. fences I think?).

rom1504:
New blocks have new properties so clearly it would change per version yes

rom1504:
Stale, re open if you want to finish it

PR #835, opened 2024-01-14
Add New Data Version - Doc update

wgaylord:
I still need to add a doc about the soon to be added `sounds.json` file. But I think this should be good or at least a bit better. 

Also want to expand out what tests to run and that sort of thing before merging this, in addition, I need to document the loginPacket and how to get it and also how to add the new version to all the files that need it defined. 

I plan on adding all the info in the next few days.


doc/add-data-new-version.md
```diff
@@ -4,46 +4,44 @@ For bedrock edition see [bedrock.md](bedrock.md)
...
+| biomes.json | Yes | Use [minecraft-data-generator-server][2] |
+| instruments | Yes | Use [minecraft-data-generator-server][2] |
+| materials.json | Yes | Use [minecraft-data-generator-server][2] |
+| windows.json | No | Look at [wiki.vg inventories][4] |
+| version.json | No | Look at [Protocol Versions][5] | [wiki.vg protocol numbers][6] |
+| effects.json | Yes | Use [minecraft-data-generator-server][2] |
+| enchantments.json | Yes  | Use [minecraft-data-generator-server][2] |
+| language.json | Yes | Use [minecraft-data-generator-server][2] |
+| particles.json | Yes | Use [minecraft-data-generator-server][2] |
+| blockLoot.json | No | Use [minecraft-jar-extractor][7] | Currently no extractor seems to works for this
```
rom1504, line 23: I don't understand why you say both that people should use Minecraft jar extractor and that there is no extractor

wgaylord: I link it since its the last know working extractor. I should probably mention that some where.

doc/add-data-new-version.md
```diff
@@ -4,46 +4,44 @@ For bedrock edition see [bedrock.md](bedrock.md)
...
-| materials.json | No | [import from previous version](18) or [fix it properly](19) |
-| windows.json | No | Look at [wiki.vg inventories][7] |
-| version.json | No | Look at [Protocol Versions][9] | [wiki.vg protocol numbers][8] |
-| effects.json | No | Look at [minecraft wiki effects][12] |
-| enchantments.json | No | Duplicate enchantments.json from the latest version and add manually the missing enchantments. Enchantments data could be found by looking into the deobfuscated classe files of each enchantments, as well as in the Enchantments registry. |
-| language.json | Yes | Use [minecraft-jar-extractor][10] |
-| particles.json | Yes | Use [Burger][1], then use [burger-extractor][2] |
-| blockLoot.json | Yes | Use [minecraft-jar-extractor][10] |
-| entityLoot.json | Yes | Use [minecraft-jar-extractor][10] |
-| mapIcons.json | No | Icons can be found in the Minecraft jar file where they are added as a single sprite. The file location is `/assets/minecraft/textures/map/map_icons.png`. Alternatively you might be able to look up the icons from the following page on the [Minecraft wiki][15] or from [wiki.vg][16]. | [minecraft-data pr mapIcons][14] |
```
extremeheat, line 25: Why remove the file path for this, is it incorrect?

wgaylord: Its a correct file path but not useful. Since it doesn't label the icons at all, and doesn't provide the information needed to fill out the mapIcons.json file.

wgaylord: For example https://github.com/extremeheat/extracted_minecraft_data/blob/client1.20.4/client/net/minecraft/world/level/saveddata/maps/MapDecoration.java  vs the map image file. https://github.com/PrismarineJS/minecraft-assets/blob/master/data/1.20.2/map/map_icons.png

doc/add-data-new-version.md
wgaylord: I have removed the mention of it. Since it doesn't work.

wgaylord:
Any further comments?

doc/add-data-new-version.md
```diff
@@ -4,46 +4,42 @@ For bedrock edition see [bedrock.md](bedrock.md)
...
+| biomes.json | Yes | Use [minecraft-data-generator-server][2] |
+| instruments | Yes | Use [minecraft-data-generator-server][2] |
+| materials.json | Yes | Use [minecraft-data-generator-server][2] |
+| windows.json | No | Look at [wiki.vg inventories][4] |
+| version.json | No | Look at [Protocol Versions][5] | [wiki.vg protocol numbers][6] |
+| effects.json | Yes | Use [minecraft-data-generator-server][2] |
+| enchantments.json | Yes  | Use [minecraft-data-generator-server][2] |
+| language.json | Yes | Use [minecraft-data-generator-server][2] |
+| particles.json | Yes | Use [minecraft-data-generator-server][2] |
+| blockLoot.json | No | Currently no extractor seems to works for this |
```
rom1504, line 23: minecraft-jar-extractor should be mentioned even if it needs updates ; until we have something better

wgaylord: Problem is minecraft-jar-extractor doesn't work at all for anything on current versions of MC.  That's why we have been using blockLoot.json and entityLoot.json from 1.17 on every version since. (At least that is my guess why.)

I only removed it since you said to not list it while also saying no extractor currently works.

wgaylord: Okay, just did some testing, seems the extractor will work for 1.20.3 but not 1.20.4, 1.20.2 or 1.20 it self which makes no sense at all (Throwing timeouts errors on getting the jars or data from mojang). So I will update this to say to use it but that it may have require fixing per versions.

Issue #833, opened 2024-01-10
Add blocks & items pc renames data


zardoy:
e.g. grass was renamed again in 1.20.3. Would be useful to track such data renames for compatibility (e.g. for some external data that is bound to specific version). Not possible to extract automatically (only validate), will add it soon. The only known version where this data cannot be mapped correctly is 1.13 where many blocks & items were merged (and there is another mapping for that), other versions have just [name changes](https://github.com/zardoy/space-squid/blob/b0707ada367549ce6efd23472b4fb9f28e384d96/src/itemBlockRenames.json#L608) afaik

Maybe would also be useful for any other data?

wgaylord:
Well you could possibly extract it from the datafixers in the game. But I am not sure this is something worth adding to minecraft-data itself. imo.

PR #836, opened 2024-01-17
add data renames (blocks & items)

zardoy:
fixes https://github.com/PrismarineJS/minecraft-data/issues/833


rom1504:
how did you generate it ?

zardoy:
For most versions I specified it manually (or almost manually by looking at data fixer code) since there are not many renames. But for the 1.13 version I used this script:

```js
//@ts-check
import fs from 'fs'
import minecraftData from 'minecraft-data'

const mcData = minecraftData('1.12.2')

const file = JSON.parse(fs.readFileSync('node_modules/minecraft-data/minecraft-data/data/pc/common/legacy.json', 'utf8'))

let map = []
for (const [key, item] of Object.entries(file.blocks)) {
    const item2 = mcData.blocks[key.split(':')[0]];
    if (!item2) continue
    const oldName = item2.name
    const newName = item.split(':')[1].split('[')[0];
    if (map.find(x => x[1] === newName && x[0] === oldName)) {
        continue
    }
    if (newName === oldName) {
        continue
    }
    map.push([oldName, newName])
}

fs.writeFileSync('map.json', JSON.stringify(map, null, 2))

```

zardoy:
Im not sure if thats possible to generate with the server. I tried to extract using text manipulation (by looking at `createRenamer` in DataFixers, but I couldn't get all renames in that way). However, does the current format / file look good?

zardoy:
Waiting for this pr, then add a function to use it to node wrapper

rom1504:
I'm looking at the file and actually I don't understand what this is.

What is a new name ? What is an old name ?

zardoy:
> What is a new name ? What is an old name ?

Also, the code above mentions the process of creating the map for 1.13: `map.push([oldName, newName])`. Why array? Because you might also want to go from a new name to an old name (like also mentioned in docs).

I will also open a pr in nmd so you can better understand how real code works with this format.

rom1504:
https://github.com/PrismarineJS/node-minecraft-data/pull/335#issuecomment-1902543462 following of the conversation there.

TLDR: most of these "renames" are not renames but actually just some block being removed in old version and some block being added in new version.

A better solution may be to create "group of similar blocks" like grass, chest,... That you could handle in a similar way in downstream code

rom1504:
The renames in this PR still contain many renames between blocks that do not have the same semantic

rom1504:
Re open if you want to finish it

Issue #820, opened 2024-01-04
Add sound id  - sound name map for each version


zardoy:
sound ids (those are used in sound packet) are completely different between versions, however this is not a case with sound names. Anyway it’s always much easier to work with sound names, so any data related to sound ids would be really handy here

I don’t think it’s related to https://github.com/PrismarineJS/minecraft-data/issues/77 which should contain names instead

zardoy:
https://pokechu22.github.io/Burger/1.19.2.html#sounds see fields. each sound has sound id and name (didn't know its called variant id). Also I noticed that mapping for some ids seems incorrect here

zardoy:
>

> Also I noticed that mapping for some ids seems incorrect here

Nevermind its just a sound_effect packet incorrectly parsed here on 1.19.2 and above

wgaylord:
Ah this will work well for 1.19.2 and lower. Since 1.19.3 and up you don't need the numerical ID to actually play the sound, you can use its full Identifier.   But having it allows you to send sounds with smaller packets. 

So still good data to have.

wgaylord:
btw are you in the discord, as that would be a nicer why to discuss all these additions.

zardoy:
> Since 1.19.3 and up you don't need the numerical ID to actually play the sound

hm, what do you mean? i still receive `hardcodedSoundEffectHeard` with numerical sound ID from most events of the world (sound to play) on 1.19.2 and above

> btw are you in the discord, as that would be a nicer why to discuss all these additions.

yes

wgaylord:
https://github.com/PrismarineJS/minecraft-data-generator-server/pull/17 Will have to make a schema here for the tests but this should work.  Also I was thinking more of as a server you can send the text ids, forgot mojang would probably always use the numerical for builtin sounds.

zardoy:
> Also I was thinking more of as a server you can send the text ids

WDYM? server sends numerical ids

wgaylord:
The Mojang server does (atleast for sounds that are built into the game), its perfectly valid for a server to send an ID of 0 and fill in the other information of a packet instead of using the IDs.

zardoy:
can be closed now (since it seems to be automated now)?

PR #839, opened 2024-01-25
Add sounds files for majority of versions.

wgaylord:
This adds the Sound ID to Name mapping information requested in #820 which are only needed for 1.9+

This has that information for the majority of the latest of all versions or the newest that the extractor would run on without much trouble. (Having to rewire other parts of the extractor)

Some versions have the lowest and highest minor versions number since the data-extractor was setup for the lowest, and I also wanted to extract the highest version.

The versions are 1.9.4, 1.10.2, 1.11.2, 1.12.2, 1.13.2, 1.14, 1.144, 1.15, 1.15.2, 1.16 (1.16.5 using 1.16 file), 1.17 (1.17.1 using 1.17 file), 1.18 (1.18.2 using 1.18 file), 1.19, 1.19.2, (1.19.4 using 1.19.2 file), 1.20.1 (1.20 using 1.20.1 file), 1.20.2, 1.20.4

This will also most likely require a PR to node-minecraft-data to provide a nice way of accessing this data.


wgaylord:
Changed to WIP since I have to go thru and check all the different versions, since for example on 1.13 the code uses the exact id out of the registry over the network, while in 1.19 for example that id gets a 1 added to it, which I did not know about.

wgaylord:
Okay looks like 1.19.2+ needs the ID in the registry offset by one in the packet. (looked thru the source of very version 1.9 to 1.20.4 to check.) Every update we should check the source to make sure we still need the offset.

rom1504:
lgtm

PR #443, opened 2021-08-02
Add 1.6.4 protocol.json

nickelpro:
Adding this with the idea that if one pre-netty protocol manages to get support, adding more later will be less of an issue.

Documentation for reviewing this commit:
* [Protocol](https://wiki.vg/index.php?title=Protocol&oldid=4899)
* [Data Types](https://wiki.vg/index.php?title=Data_types&oldid=4806)
* [Metadata](https://wiki.vg/index.php?title=Entity_metadata&oldid=4900)
* [Slot Data](https://wiki.vg/index.php?title=Slot_Data&oldid=4753)
* [Object Data](https://wiki.vg/index.php?title=Object_Data&oldid=4754)

Note that this doesn't add support to NMP for 1.6.4 due to two issues:
* No protocol states
* Strings are UTF-16

The latter is something of an issue, since MCD lacks the ability to signify the change in string encoding.

PR #840, opened 2024-01-26
1.6.4 Protocol

brenoxavier:
This is my attempt to continue the work started by this PR #443.

I've written the specifications for protocol 78 (pre Netty) in version 1.6.4, but I have some doubts about whether this is correct because some packages are quite different from the post-Netty versions.

By setting the strings to UTF-16, I found that the protodef-validator is failing to validate the latest version of ProtoDef.

Used documentation:

[Protocol](https://wiki.vg/index.php?title=Protocol&oldid=4899)
[Data Types](https://wiki.vg/index.php?title=Data_types&oldid=4806)
[Metadata](https://wiki.vg/index.php?title=Entity_metadata&oldid=4900)
[Slot Data](https://wiki.vg/index.php?title=Slot_Data&oldid=4753)
[Object Data](https://wiki.vg/index.php?title=Object_Data&oldid=4754)


wgaylord:
It would be best to fix the tests so that they pass.

wgaylord:
Looks like an issue here https://github.com/ProtoDef-io/ProtoDef/blob/master/schemas/utils.json Where pstring only takes the count type and not encoding in the schema.  I have opened an issue on the ProtoDef repo for it. https://github.com/ProtoDef-io/ProtoDef/issues/51

wgaylord:
Looks like https://github.com/ProtoDef-io/node-protodef-validator/pull/11 will fix this.

extremeheat:
Can you be specific here with 1.6.4 instead of 1.6? Or were there no changes in protocol between then?

brenoxavier:
> Can you be specific here with 1.6.4 instead of 1.6? Or were there no changes in protocol between then?

I couldn't identify major differences in the protocol. I believe the protocol from 1.6.4 should work for all versions of 1.6.

https://wiki.vg/index.php?title=Protocol&type=revision&diff=4657&oldid=1096

extremeheat:
There should be any ambiguity there, if there is a protocol version change then there likely is a difference.

data/pc/1.6/version.json
extremeheat, line 1: This needs to be added to the protocolVersions.json file

brenoxavier:
> There should not be any ambiguity there, if there is a protocol version change then there likely is a difference.

Ok, I'll specify it as 1.6.4.

data/pc/1.6/version.json
brenoxavier: The version is already present in the file, does it need any other adjustments?

image

extremeheat:
What version did you use as a base to writing this?

brenoxavier:
> What version did you use as a base to writing this?

I used 1.7

extremeheat:
Yeah, I'm diffing 1.6.4 protocol.json to 1.7 and it seems there are alot of differences. Did the protocol really change that much or did you re-organize things? That would make review much harder as would need to go through all the packets individually

wgaylord:
> Yeah, I'm diffing 1.6.4 protocol.json to 1.7 and it seems there are alot of differences. Did the protocol really change that much or did you re-organize things? That would make review much harder as would need to go through all the packets individually

Yes, because 1.6 4 is pre-netty.

brenoxavier:
> Yeah, I'm diffing 1.6.4 protocol.json to 1.7 and it seems there are alot of differences. Did the protocol really change that much or did you re-organize things? That would make review much harder as would need to go through all the packets individually

Many things changed from 1.6 to 1.7, I tried to keep them as close as possible.

extremeheat:
I see. So yes this would require a more thorough review. Since there is no concept of states or clientbound/serverbound packets, it would probably be cleaner to not use the 1.7 protocol as reference (as its structure assumes those things, like toServer/toClient/play/etc).

1.7.6-1.7.10 protocol - https://wiki.vg/index.php?title=Protocol&oldid=6003
1.6.4 protocol - https://wiki.vg/index.php?title=Protocol&oldid=4899

brenoxavier:
So is it better to specify packages in just a single structure?

brenoxavier:
> I see. So yes this would require a more thorough review. Since there is no concept of states or clientbound/serverbound packets, it would probably be cleaner to not use the 1.7 protocol as reference (as its structure assumes those things, like toServer/toClient/play/etc).
> 
> 1.7.6-1.7.10 protocol - https://wiki.vg/index.php?title=Protocol&oldid=6003
> 1.6.4 protocol - https://wiki.vg/index.php?title=Protocol&oldid=4899

So is it better to specify packages in just a single structure?

extremeheat:
The point of that separation is so node-Minecraft-protocol can swap serializers when the state changes. That's not necessary for 1.6, so there should be one mapper for switching between packets (in the single state).

brenoxavier:
> The point of that separation is so node-Minecraft-protocol can swap serializers when the state changes. That's not necessary for 1.6, so there should be one mapper for switching between packets (in the single state). It technically could be separate in the data here but it doesn't make sense because the implementation in nmp would have to dedupe the switches.

Would a structure like this be a good idea? Segregate only into client packages, server packages, and packages used by both.

`{
  "types": {

  },
  "toClient": {

  },
  "toServer": {

  },
  "twoWay": {

  }
}`

extremeheat:
The point of protocol.json is something that protodef can read and use inside node-minecraft-protocol. It's not designed around documenting anything for sake of it (like what direction the packets are). It just needs to work inside node-minecraft-protocol. You can make an accompanying PR to node-minecraft-protocol if you think you can make it work that way. What I'm saying is I don't see how that will work without special handling for merging the main packet switch, at least not in nice way in the code, as it would require some preprocessing to make it work your way.

brenoxavier:
> The point of protocol.json is something that protodef can read and use inside node-minecraft-protocol. It's not designed around documenting anything for sake of it (like what direction the packets are). It just needs to work inside node-minecraft-protocol. You can make an accompanying PR to node-minecraft-protocol if you think you can make it work that way. What I'm saying is I don't see how that will work without special handling for merging the main packet switch, at least not in nice way in the code, as it would require some preprocessing to make it work your way.
> 
> What I meant was you just need { types, packets: {types} } and in node-minecraft-protocol you would do .addTypes(protocol.types); .addTypes(protocol.packets.types) to utilize it. Since there would be no collisions there would be no problem that way. Or just put them all directly into the root types.

Perfect, I'll simplify the packet structure and work to make node-minecraft-protocol compatible with it.

wgaylord:
I see some large issues so far. Why are their varints anywhere in the protocol.json. 1.6.4 didn't use varints at all. 

Multiple packets don't line up even with Wiki.vg for 1.6.4 Example the packet_encryption_key_response packet you have the handling of the byte arrays all incorrect where you read the length but then tell the array that it need to read the length as a byte itself. (I think this might be from now knowing how to use ProtoDef to do that more then just errors.)  You also have this multiple times in the file?

packet_encryption_key_request - Doesn't match wiki.vg at all 

We may want to change some field names on packet_player_position_and_look due to the order changing between if the packet is going from server to client or client to server

packet_player_digging - Missing the status byte as the first argument

packet_bed - Missing a byte in it.

And those are only some of the issues I found at first glance, let alone the organizational issue of 1.6.4 not having server vs client or the concepts of states.

brenoxavier:
> I see some large issues so far. Why are their varints anywhere in the protocol.json. 1.6.4 didn't use varints at all. 
> 
> Multiple packets don't line up even with Wiki.vg for 1.6.4 Example the packet_encryption_key_response packet you have the handling of the byte arrays all incorrect where you read the length but then tell the array that it need to read the length as a byte itself. (I think this might be from now knowing how to use ProtoDef to do that more then just errors.)  You also have this multiple times in the file?
> 
> packet_encryption_key_request - Doesn't match wiki.vg at all 
> 
> We may want to change some field names on packet_player_position_and_look due to the order changing between if the packet is going from server to client or client to server
> 
> packet_player_digging - Missing the status byte as the first argument
> 
> packet_bed - Missing a byte in it.
> 
> And those are only some of the issues I found at first glance, let alone the organizational issue of 1.6.4 not having server vs client or the concepts of states. 
> 
> Also we may want to rename some packets to be consistent with the other versions, provided they do the same type of thing.

I will review the packages and fix this. This is my first contact with ProtoDef.

brenoxavier:
I made some changes to the protocol.json:

- Simplified the protocol specification by removing states and directions. (@extremeheat)
- Reviewed all packages and corrected those identified as incorrect (@wgaylord).
- Aligned the protocol to closely resemble version 1.7.10.

wgaylord:
> * Reviewed all packages and corrected those identified as incorrect (@wgaylord).
I will go thru all the packets sometime soon and compare to source. Hopefully everything will line up well.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+                "countFor": "records"
+              }
+            ]
+          },
+          {
+            "name": "records",
+            "type": [
+              "array",
+              {
+                "count": "recordCount",
```
wgaylord, line 1350: This needs to be looked at, as in 1.6.4 these records are just positions relative to the explosion written as bytes. 

See https://wiki.vg/index.php?title=Protocol&oldid=4899#Explosion_.280x3C.29

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+          {
+            "name": "position",
+            "type": "position_iii"
+          },
+          {
+            "name": "volume",
+            "type": "f32"
+          },
+          {
+            "name": "pitch",
```
wgaylord, line 1450: This should be an unsigned byte `u8`

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+          {
+            "name": "slotCount",
+            "type": "i8"
+          },
+          {
+            "name": "useProvidedTitle",
+            "type": "bool"
+          },
+          {
+            "name": "entityId",
```
wgaylord, line 1550: This is conditional on inventoryType equaling 11 according to the source other wise this field doesn't exist

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+            "name": "line3",
+            "type": "string"
+          },
+          {
+            "name": "line4",
+            "type": "string"
+          }
+        ]
+      ],
+      "packet_item_data": [
```
wgaylord, line 1714: I think this should be renamed since its only used for sending over the images for on Maps.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+          {
+            "name": "itemType",
+            "type": "i16"
+          },
+          {
+            "name": "itemId",
+            "type": "i16"
+          },
+          {
+            "name": "text",
```
wgaylord, line 1726: This is just raw data in the form of a byte array of the length its prefixed with.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+          {
+            "name": "itemName",
+            "type": "string"
+          },
+          {
+            "name": "action",
+            "type": "i8"
+          },
+          {
+            "name": "scoreName",
```
wgaylord, line 1885: scoreName and value are are not read or written if action = 1 

See 1.6.4 source or 
https://wiki.vg/index.php?title=Protocol&oldid=4899#Update_Score_.280xCF.29 for more details.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+            "type": [
+              "buffer",
+              {
+                "countType": "i16"
+              }
+            ]
+          }
+        ]
+      ],
+      "packet_legacy_server_list_ping": [
```
wgaylord, line 2077: This should just be packet_server_list_ping the server responds using a kick packet.

It has the following fields seen in source and at https://wiki.vg/Server_List_Ping#1.6

a byte that just always holds 1
a byte that for some reason holds the ID of the CustomPayloadPacket
a string - This is just a constanct in this case its always "MC|PingHost"
a short  Length of the rest of the data in the packet
a byte - protocol version
a string  - host name of the server
a int - port of the server

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+                    "name": "itemCount",
+                    "type": "i8"
+                  },
+                  {
+                    "name": "itemDamage",
+                    "type": "i16"
+                  },
+                  {
+                    "name": "nbtData",
+                    "type": "compressedNbt"
```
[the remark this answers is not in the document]
brenoxavier: It's the same as version 1.7; I believe it should work.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+            "name": "yaw",
+            "type": "i8"
+          },
+          {
+            "name": "objectData",
+            "type": [
+              "container",
+              [
+                {
+                  "name": "intField",
```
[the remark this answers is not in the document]
brenoxavier: Should it be renamed to something like "entityId"?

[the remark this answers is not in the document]
wgaylord: maybe ownerId?  Since its not the ID of the entity but the player that placed / spawned it.

data/pc/1.6.4/protocol.json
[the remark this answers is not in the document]
wgaylord: This does seem correct, just decompiled 1.7 to check and while it uses varint all over it still uses a short for the NBT Data length.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2290 @@
...
+            "type": [
+              "buffer",
+              {
+                "countType": "i16"
+              }
+            ]
+          }
+        ]
+      ],
+      "packet_server_list_ping": [
```
wgaylord, line 2081: Still have the issue that this packet is not implemented at all here.

data/pc/1.6.4/protocol.json
brenoxavier: Should I remove the field or add a conditional?

wgaylord: It should be conditional using a switch I believe.

data/pc/1.6.4/protocol.json
brenoxavier: Something like "packet_map_data"?

wgaylord: 1.7+ uses just `packet_map` for the packet that sends the map items image, so that is probably the best name for it.

data/pc/1.6.4/protocol.json
brenoxavier: How would the specification for this be?

wgaylord: It would look something like this https://github.com/PrismarineJS/minecraft-data/tree/master/data/pc/1.7/protocol.json#L2042-L2068

data/pc/1.6.4/protocol.json
brenoxavier: How should I proceed?

wgaylord: I documented it here, https://github.com/PrismarineJS/minecraft-data/pull/840#discussion_r1473553320 although the best way would be to open up 1.6.4 source and do it following that.

data/pc/1.6.4/protocol.json
```diff
@@ -0,0 +1,2286 @@
...
+      {
+        "compareTo": "$compareTo",
+        "fields": {
+          "0": "i8",
+          "1": "i16",
+          "2": "i32",
+          "3": "f32",
+          "4": "string",
+          "5": "slot",
+          "6": [
```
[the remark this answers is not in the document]
brenoxavier: I couldn't quite understand how this should be done; in the other versions, they are bitfields.

rom1504:
hi @brenoxavier would you like to try and open a PR on nmp using this ? 
otherwise it's a bit hard to tell if this would work

brenoxavier:
> hi @brenoxavier would you like to try and open a PR on nmp using this ? otherwise it's a bit hard to tell if this would work

It won't work. I will analyze how to modify the nmp to make it work with this.

rom1504:
ok this is old; not going to happen

feel free to ask to re open or open a new pr if you want

PR #853, opened 2024-02-26
Fixed 1.20.3 and 1.20.4 loginPacket version was set incorrectly on da…

n-qber:
In data/dataPaths.json the loginPacket version was set to 1.19.2, but changes were made in 1.20.2 what lead to issues when using loginPacket for 1.20.3 and 1.20.4, updating them in data/dataPaths.json solves the issues

### DIFF
--- a/data/dataPaths.json
+++ b/data/dataPaths.json
@@ -1294,7 +1294,7 @@
       "particles": "pc/1.20.3",
       "blockLoot": "pc/1.20",
       "entityLoot": "pc/1.20",
-      "loginPacket": "pc/1.19.2",
+      "loginPacket": "pc/1.20.2",
       "tints": "pc/1.20.2",
       "mapIcons": "pc/1.20.2",
       "commands": "pc/1.20.3"
@@ -1319,7 +1319,7 @@
       "particles": "pc/1.20.3",
       "blockLoot": "pc/1.20",
       "entityLoot": "pc/1.20",
-      "loginPacket": "pc/1.19.2",
+      "loginPacket": "pc/1.20.2",
       "tints": "pc/1.20.2",
       "mapIcons": "pc/1.20.2",
       "commands": "pc/1.20.3"


wgaylord:
Would be better to get a 1.20.3 / 1.20.4 logging packet from nmp then use the 1.20.2 packet.

n-qber:
Thanks for the reply, while trying to execute that, I fond out the problem was not on the login event, but rather the "registry_data" event sent by the server:

The code below show the line where registry_data is sent, note that it relies on options.registryCodec
https://github.com/PrismarineJS/node-minecraft-protocol/blob/e50b604a60403e7412e5124f29cc31a2c00d66ab/src/server/login.js#L214

Note, in the code below, that options.registryCodec comes from either _options, mcData or mcData.loginPacket_
https://github.com/PrismarineJS/node-minecraft-protocol/blob/e50b604a60403e7412e5124f29cc31a2c00d66ab/src/createServer.js#L49

I manually checked and for all versions noted here: _1.20.2, 1.20.3 and 1.20.4_ the value for mcData.registryCodec is indeed undefined, which means that the value (options.registryCodec) used gets set to mcData.loginPacket.dimensionCodec.

So to implement your idea I had to set the option registryCodec to the updated version (1.20.2) so that it doesn't use the (1.19.2) version:

```javascript
const mc = require('minecraft-protocol')
const nbt = require('prismarine-nbt')


const mcData =        require('minecraft-data')('1.20.4') // VERSION BEING USED (ends up using loginPacket 1.19.2 version)
const updatedMcData = require('minecraft-data')('1.20.2') // VERSION THAT SHOULD BE USED (uses 1.20.2 loginPacket version)

console.log(mcData.registryCodec)
console.log(updatedMcData.registryCodec)

const noOptionServer = mc.createServer({
  port: 25565,
  version: '1.20.4',
  motd: "No option set"
});

const currentOptionServer = mc.createServer({
  port: 25566,           // optional
  version: '1.20.4',
  motd: "Current Option Set",
  registryCodec: mcData.loginPacket.dimensionCodec
});

// noOptionServer and currentOptionServer will use the same registryCodec, because of what I
// mentioned earlier

const updatedOptionServer = mc.createServer({
  port: 25567,           // optional
  version: '1.20.4',
  motd: "Updated Option",
  registryCodec: updatedMcData.loginPacket.dimensionCodec
});


// Note that the first two severs, generate errors on the client, while the last one allows the login process
// In this example they don't generate a world so the client will wait on "Joining world..." screen forever
```

wgaylord:
Still the correct way to do this would be the get the Login Packet for 1.20.4 from node-minecraft-protocol's tests for 1.20.4

extremeheat:
I agree there should be a line at https://github.com/PrismarineJS/node-minecraft-protocol/blob/e50b604a60403e7412e5124f29cc31a2c00d66ab/src/createServer.js#L49 explaining what's happening.

First, the user can specify a custom dimensionCodec within the options. Or, if the dimension codec exists inside minecraft-data exported as .registryCodec, then that will be used. We currently don't expose this information like this at the moment, but this could be done in a future PR.

Finally, the method gets the dimension codec from the login packet. This is how it's done in versions prior to 1.20.*. For compatibility with existing code, we just keep the dimension codec inside the login packet when generating it. The intended way to generate the loginPacket.json per the [doc](https://github.com/PrismarineJS/minecraft-data/blob/master/doc/add-data-new-version.md) is to run the tests inside node-minecraft-protocol, which will automatically generate loginPacket.json's for all the versions tested.

So again, just run `npm test` inside nmp to get these files.

rom1504:
this is better than master state; merging

PR #864, opened 2024-03-22
Bedrock 1.19.80-1.20.71 blocks data

FreezeEngine:
- 1.20.50-1.20.71 should be close to perfect
- 1.19.80-1.20.40 blockStates and blocks tested but other data not sure

Data from extractor + used "checkout" to rollback some repos to generate older data


extremeheat:
Would be great if there was a way to test this data

FreezeEngine:
> Would be great if there was a way to test this data

Block states were generated from geyser palletes, I have ran a small test on first state of almost each block, I think I can publish it soon (basically takes block names from block enum in avaliable commands packet, runs /setblock foraech of them and awaits for block to change -> gets block name from blocksByStateId and compares)

Also might be able to implement it as a test here

.60 and .70 data I tested with my bot by looking if it decoded pallete correctly and if it fell through the world, collisions were always strange on some blocks, seems like state to correct bounding box is not working propperly (mostly feels like every state has only one collision or is rotated wrongly except for default state)

extremeheat:
Yes, right now there is no way to validate this data, but there should be. The server sends a block palette checksum so we could check against that in the future. But ideally this is all automated without needing manual work.

extremeheat:
In the meantime though I'll merge to not create a PR backlog, we can figure out the validation later

PR #869, opened 2024-04-01
Add blocks.json to 1.20.71

AnyBananaGAME:
Blocks change in every version and not everyone knows and has a way of dumping them so ill pr em


AnyBananaGAME:
It appears that it may be

AnyBananaGAME:
compared the state ids and they look same to me

PR #870, opened 2024-04-16
Fixed 1.20.71 correctPlayerMovePredictionPacket missing `prediction_type`

Cdm2883:


extremeheat (approved):

extremeheat:
Can you apply this fix to previous versions where it's also applicable?

Cdm2883:
> Can you apply this fix to previous versions where it's also applicable?

The update to this packet structure seems to start in protocol 662, only 1.20.70+

extremeheat:
Where are you seeing that? Seems like it should be applied since 1.20.40 (protocol v622)

https://github.com/Mojang/bedrock-protocol-docs/blob/main/html/svg/CorrectPlayerMovePredictionPacket.svg

rom1504:
should we merge this @extremeheat ?

Cdm2883:
> should we merge this @extremeheat ?

I don't think this should be done at this time. I haven't freed up time to check previous versions lately, and then there's the fact that I've found more wrong in packets, and I'd like to continue fixing other packets on this PR :D

Should the status of this PR go to draft or something like that?

rom1504:
Up to you about the status, but thanks for clarifying!

PR #876, opened 2024-05-07
1.18+ Book editing feature

unlimitedcoder2:
In 1.18 a new book editing packet was added
https://github.com/prismarinejs/mineflayer/issues/3204

### DIFF
--- a/data/pc/common/features.json
+++ b/data/pc/common/features.json
@@ -238,7 +238,7 @@
   },
   {
     "name": "hasEditBookPacket",
-    "description": "book editing is handled through a packet",
+    "description": "book editing is handled through a packet with the updated book item",
     "versions": [
       "1.13",
 	  "1.17.1"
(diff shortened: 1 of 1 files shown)


data/pc/common/features.json
```diff
@@ -238,9 +238,17 @@
...
     "name": "hasEditBookPacket",
-    "description": "book editing is handled through a packet",
+    "description": "book editing is handled through a packet with the updated book item",
     "versions": [
       "1.13",
+	  "1.17.1"
+    ]
+  },
+  {
+    "name": "hasNonNbtEditBookPacket",
```
extremeheat, line 248: The feature naming is confusing here. Can you elaborate what exactly the difference in packet is before vs after?

unlimitedcoder2: The old edit book packet serializes the item and uses the nbt data from that
image

New packet sends the pages and title as packet fields
image

extremeheat: How about adding a `editBookPacketUsesNbt` feature then? This seems clearer than the current naming between `hasEditBookPacket` and `hasNonNbtEditBookPacket`, as there is actaully an edit book packet on both version ranges.

extremeheat: Done

extremeheat (approved):

PR #878, opened 2024-06-03
1.20.5 data

qwqtoday:


extremeheat:
You can remove the protocol data from the PR and add it in a different PR. That will require more manual review.

qwqtoday:
Ok

qwqtoday:
I made another pr so the commit history will be nicer #879

PR #879, opened 2024-06-09
Add some 1.20.5 data

qwqtoday:
1.20.5 update


data/pc/1.20.5/foods.json
```diff
@@ -0,0 +1 @@
+[]
```
extremeheat, line 1: is this supposed to be blank?

data/pc/1.20.5/entities.json
```diff
@@ -0,0 +1,1302 @@
+[
```
extremeheat, line 1: ok for now, but there is additional data to extract here with the extractor in tools/js/extractMcpcEntityMetadata for extracting metadata keys (used in mineflayer)

data/pc/1.20.5/foods.json
extremeheat: Likely a failure of the extractor, this would need to be fixed. Doesn't seem like there was any changes to 1.20.5 food list so it may be possible to write a script to just update the item IDs.

tools/js/extractPcFoods.js
```diff
@@ -0,0 +1,58 @@
+const fs = require('fs')
```
extremeheat, line 1: I fixed the foods for now with a new extraction script here via the source code

extremeheat:
LGTM @rom1504

extremeheat:
Going to merge this to not conflict with/block 1.20.5 protocol data

rom1504:
Thanks for handling it @extremeheat

PR #885, opened 2024-06-23
Implement protodef-yaml for mcpc

extremeheat:
* Implement protodef-yaml for mcpc
* support pc in  incrementVersion script


extremeheat:
Closing for #886 so 1.20.5 changes can be applied against it

PR #886, opened 2024-06-23
Implement protodef-yaml for mcpc

extremeheat:
* Implement protodef-yaml for mcpc
* Support pc in incrementVersion script


rom1504:
it does look pretty good

do you think we can do anything about older versions?

rom1504:
I think we should update docs for explaining contributors where they need to update the protocol info

rom1504:
sorry for the delay in reviewing, I've been pretty busy in last months.

This looks pretty good!

One last thing I am noticing is the addition of default void in many places; which seems incorrect to me. For most switches the default is not void. There is no default! In protodef it means it will fail if value is not among the specified choices.
Could you fix that?

rom1504 (approved):
Thank you for this large improvement!

PR #887, opened 2024-06-23
Add pc 1.20.5 protocol data

extremeheat:


extremeheat:
Removed feature
```json
{
    "name": "attributeIdsAreInts",
    "description": "Update Attributes packet now has integer attribute prop ids in `key`",
    "versions": ["1.20.5", "latest"]
  },
  ```
in favor of mapping the varint to a string in protocol as we don't have an attributes data extractor (required for mapping ids=>string)

File: data/pc/latest/proto.yml [hunk 1]
```diff
@@ -1,4 +1,4 @@
-!version: 1.20.3
+!version: 1.20.5
```
rom1504, line 1: that latest file has the nice benefit of having a readable diff for this kind of pr

rom1504:
can you re-open this on main ?

PR #944, opened 2024-11-30
Bedrock serverbound_loading_screen malformed packet

JSbETms:
When the `serverbound_loading_screen` is sent serverbound, the client can be kicked by the server due to a "malformed" packet with the specified paramaters.
```
{
  violation_type: 'malformed',
  severity: 'terminating',
  packet_id: 312,
  reason: 'BinaryStream read() overflow checkedNumber = 8, mReadPointer = 4, buffer Length is = 5 bytes\n' +
    'readNoHeader failed! packetId: 312'
}
```
This can be triggered by changing between dimensions using a nether or end portal in a relay server using the [docs/API](https://github.com/Sandertv/gophertunnel/blob/master/minecraft/protocol/packet/server_bound_loading_screen.go) code within bedrock-protocol.
[Here](https://github.com/Sandertv/gophertunnel/blob/master/minecraft/protocol/packet/server_bound_loading_screen.go) is the refrence I have used for the changes made.

### DIFF
--- a/data/bedrock/1.21.20/proto.yml
+++ b/data/bedrock/1.21.20/proto.yml
@@ -4227,7 +4227,7 @@ packet_serverbound_loading_screen:
    !bound: server
    # The type of the loading screen event.
    type: zigzag32
-   loading_screen_id?: varint
+   loading_screen_id?: lu32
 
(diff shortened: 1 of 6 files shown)


extremeheat (approved):

extremeheat:
Thanks

PR #948, opened 2024-12-05
Pc protocol updates

extremeheat:


rom1504:
We have protodef in now
Next finish that and nmp

rom1504:
will unblock https://github.com/PrismarineJS/flying-squid/issues/665 and https://github.com/PrismarineJS/flying-squid/issues/664

extremeheat:
If anyone wants to help contribute the process here is not hard. Just need to backwards diff all the versions's proto.yml files and apply any missing fixes to the old version.

Click . in the GitHub page for the branch to open vscode and find 1.11/proto.yml and compare with 1.10/proto.yml (right click a file and do "select for compare..." and "compare with selected), then make fixes

Context- many times when updating the protocol people fix the data for the a new version (fix because the data was always wrong in mc-data, not that mc code changed) but they don't apply the changes to old versions. So this introduces the need to use features and pushes other ugly logic to everything else. Eg, 1.21 updates a packet, some things are changed to the structure unrelated to 1.21 but those structure changes aren't applied backwards so you need a new "feature" to use the packet now. 

Also can use the opportunity to improve old definitions (eg using mappers instead of switching on changing numbers). 

So to implement this in mineflayer/nmp/fs, we need to go through all the packet listens and writes one by one and check if it changed in this PR.

extremeheat:
master issue - https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/latest/proto.yml#L1557 is not consistent across states (config vs play state uses varint vs mapper)

rom1504:
Idk how to mark back as draft

rom1504:
I think we should try and identity the most important issues here and try to merge those first

rom1504:
For example having correct player_info seems like a pretty good start https://github.com/PrismarineJS/minecraft-data/issues/972

data/pc/latest/proto.yml
```diff
@@ -61,136 +56,77 @@
...
+      - writable_book_content
+      - written_book_content
+      - trim
+      - debug_stick_state
+      - entity_data
+      - bucket_entity_data
+      - block_entity_data
+      - instrument
+      - ominous_bottle_amplifier
+      - jukebox_playable
```
extremeheat, line 112: Why jukebox_playable has 3 possible sound types (string, integer via holdset, container type)

### The Purpose of `Holder`

A `Holder` is a "smart pointer" or a reference to a registered object. Think of the `jukebox_song` registry. It contains entries like `minecraft:precipice`, `minecraft:creator`, etc. Each of these has a unique string ID and a corresponding numeric ID.

The `Holder` type is designed to represent one of two states:

1.  **Reference by ID:** "I want to use the song with numeric ID `5`." This is what happens when `readHolder` reads a non-zero varint. It's compact and efficient.
2.  **Inline Definition:** "I want to use a song that *isn't in the registry*. Here is its full data definition right now." This is what happens when `readHolder` reads a zero, and then it proceeds to read the full `JukeboxSong` data structure. This provides flexibility for custom content (e.g., from maps or servers) without needing to modify the client's built-in registries.

### Why `jukebox_playable` is Different

Now, why does `jukebox_playable` add that extra `boolean` and a third option?

The `jukebox_playable` component is used for music discs. The component needs to represent two different *kinds* of things:

*   **A "real" Jukebox Song:** This is an entry in the `jukebox_song` registry, like `minecraft:precipice`. These songs have defined properties: a sound event, a description, a comparator output, etc. This is the case that the `Holder<JukeboxSong>` handles.

*   **An arbitrary Sound Event:** This is for the "Disc Fragment 5", which doesn't play a full song but just the sound event `minecraft:item.music_disc.fragment_5`. This sound event is in the `sound_event` registry, *not* the `jukebox_song` registry. It doesn't have a defined description, comparator output, or song length. It's just a sound.

The protocol designers needed to handle both cases. They couldn't just use a `Holder<JukeboxSong>` because that wouldn't cover the simple "play this sound" case. They also couldn't just use a `Holder<SoundEvent>` because that wouldn't cover the full song data case.

### The Three Possibilities

So, the `boolean` `hasHolder` acts as a top-level switch to differentiate between these two fundamental types.

*   **If `hasHolder` is `true`:** The game expects a "real" jukebox song. It then uses the standard `Holder` logic:
    *   Reads a `varint`. If it's `> 0`, it's the numeric ID of a song in the `jukebox_song` registry. **This is your integer case.**
    *   If it's `0`, it's an inline, custom `JukeboxSong` definition. **This is your container case.**

*   **If `hasHolder` is `false`:** The game expects just a simple sound to play. It reads a `string`, which is the resource location of a sound in the `sound_event` registry (e.g., `"minecraft:item.music_disc.fragment_5"`). **This is your string case.**

So, you are absolutely correct! The three final possibilities are:
1.  **A string** (the resource key for a simple sound).
2.  **An integer** (the numeric ID for a registered jukebox song).
3.  **A container** (the full data for a custom, unregistered jukebox song).

It's a "tri-state" component, and using a `boolean` to switch between a `Holder` (which is itself a bi-state object) and a simple `string` was the most direct way for them to implement it. It is a bit weird, but it makes sense when you consider the two different kinds of audio a music disc item might need to represent.

extremeheat:
Diff with only protocol yaml changes -- https://gist.github.com/extremeheat/98663c8239609826301f4dc90013408d

rom1504:
Nmp and mineflayer tests pass, let's merge

PR #962, opened 2025-01-06
Remove Unnecessary Field for ItemStackRequest Optional Action and Update Steve Skin for Bedrock

yowzaoppo:
The unnecessary field removed in the Optional action is "times_crafted". It appears that it is no longer used in v1.21.50.
I have also provided a newly updated Steve skin attached to this pull request.


data/bedrock/latest/types.yml
```diff
@@ -1078,7 +1078,6 @@ ItemStackRequest:
          if optional:
             # For the cartography table, if a certain MULTI recipe is being called, this points to the network ID that was assigned.
             recipe_network_id: varint
-            times_crafted: u8
```
extremeheat, line 1081: When did this change? If it's old please apply to all the other relevant versions

yowzaoppo: I can't be exact when this was removed. It had to be either in v1.21.30 or v1.21.40 when this was removed. I've tried looking around for a answer.

yowzaoppo: I can confirm it was removed in v1.21.30. I have modified both v1.21.30 and v1.21.40 to remove the field.

extremeheat: I just checked and gophertunnel and cloudburst both have this at 1.16 (1.16.201). So it should be applied to pretty much all the versions

extremeheat: You can update everything with control+shift+F, paste in old code as query, put new code block as replacement, click update all occurrences, review, then rebuild

yowzaoppo: > I just checked and gophertunnel and cloudburst both have this at 1.16 (1.16.201). So it should be applied to pretty much all the versions

The field was added here when gophertunnel was getting updated to support v1.21.20:
https://github.com/Sandertv/gophertunnel/commit/d900285691c547aa2ef840094d993347eca3ed7a#diff-3d0d084d8e48bea087ffcc600251c0864a51c85f0e8dd3901e4b7452892e5c95R473

Then later, someone did a PR removing the field while gophertunnel was supporting v1.21.30:
https://github.com/Sandertv/gophertunnel/commit/27724f7679d7c7ab6b5ac9a3e897c2600ee3f612#diff-3d0d084d8e48bea087ffcc600251c0864a51c85f0e8dd3901e4b7452892e5c95L533

So in this case, it had to been removed in v1.21.30. None of the types data or protocol data below v1.21.20 doesn't have the times_crafted for the optional action. So the only version that has the field is v1.21.20.

Long story short, the field was added in v1.21.20, but then it had to be removed in v1.21.30. I don't know if I misunderstood you here, but I'm sure everything here is alright.

extremeheat: Ah, I see. Already not present here for <1.21.20

extremeheat:
Thanks

PR #971, opened 2025-01-09
Add skinUsesKeyedProperties

Pix3lPirat3:
`player_info` packet uses the `key` field for skin properties instead of `name`, Applies to Minecraft 1.19.3 and above.

### DIFF
--- a/data/pc/common/features.json
+++ b/data/pc/common/features.json
@@ -855,5 +855,10 @@
     "name": "noAckOnCreateSetSlotPacket",
     "description": "Server does not send ack on creative_set_slot packets",
     "versions": ["1.21.3", "latest"]
+  },
+  {
+    "name": "skinUsesKeyedProperties",
+    "description": "`player_info` packet uses the `key` field for skin properties instead of `name`",
+    "versions": ["1.19.3", "latest"]
   }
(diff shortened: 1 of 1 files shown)


extremeheat (approved):

extremeheat (changes requested):
Actually... this is something we should fix in the data instead... can be added to work in https://github.com/PrismarineJS/minecraft-data/pull/948

Pix3lPirat3:
@rom1504 approved the PR for mineflayer https://github.com/PrismarineJS/mineflayer/pull/3548 so should that be undone? If you're wanting the fix to be done elsewhere

rom1504:
The PR you mention is not approved and also not merged. Do you mean another one ?

Pix3lPirat3:
> The PR you mention is not approved and also not merged. Do you mean another one ?

I confused myself cause you fixed my messup of multiple commits

rom1504:
@Pix3lPirat3 can you fix that directly in protocol.yml ?

extremeheat:
I think this was fixed in #948. Is it still valid?

Pix3lPirat3:
Fixed by merged #948

PR #974, opened 2025-01-10
Optimize doc using chatgpt.

rom1504:
Changes
- pagination
- lazy rendering by subsets instead of prerendering everything
- some formatting

### DIFF
--- a/showValues.js
+++ b/showValues.js
@@ -19,122 +19,224 @@ function toggleAnchor () {
 }
 
 function fieldsToColumns (fields) {
-  return fields.map(function (field) { return { title: field } })
+  return fields.map(function (field) {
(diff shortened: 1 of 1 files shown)


rom1504:
it goes from 4s to 0.5s rendering

what do you think @extremeheat ?

rom1504:
looks like 
image

which I think is fine

rom1504:
ah but that breaks search and filtering...

rom1504:
fixed

rom1504:
canvas feature of chatgpt is interesting, it shows line by line edit
however it's pretty slow

extremeheat (approved):
LGTM

PR #979, opened 2025-01-27
Use deepseek to extract comments from wikivg into the protocol

rom1504:
Prompt

```
I would like you to match the packets from these 2 files (they have approximately the same names), then add the descriptions from the wikivgfirstpart.txt files as comments in the packets of the proto_without_common.yml file then send me that augmented proto_without_common.yml file. if you don't find a packet; don't invent the description and leave the proto_without_common.yml packet as is. Still try your best to find the matching packet; the case is different and the wording might be slightly different
For example packet_common_select_known_packs is Clientbound Known Packs
so 
Informs the client of which data packs are present on the server. The client is expected to respond with its own Serverbound Known Packs packet. The Notchian server does not continue with Configuration until it receives a response.

The Notchian client requires the minecraft:core pack with version 1.21.4 for a normal login sequence. This packet must be sent before the Registry Data packets.

should be added as a comment with # 
before packet_common_select_known_packs

don't be imaginative and just reuse the description from the wikivgfirstpart.txt file. Also please add the name of the wikivgfirstpart.txt packet in the yml comment prefixed by wiki.vg: 
```

and give it 2 files:
1. proto.yml (remove the non packet part at the start)
2. https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol (copy paste as text from browser)

it eventually went above the length limit, so to avoid that I'll first do the client bound then the server bound next time

### DIFF
--- a/data/pc/latest/proto.yml
+++ b/data/pc/latest/proto.yml
@@ -1274,70 +1274,95 @@
             suggestionType: ../flags/has_custom_suggestions ?
                if 1: string
                default: void
-   # # Shared Packets
-   # These are packets that are shared between multiple states.
-   # They are defined here to avoid duplication.
-   # =====
-   # MC: ClientboundCookieRequestPacket
-   packet_common_cookie_request:
-      cookie: string
-   # MC: ClientboundStoreCookiePacket
-   packet_common_store_cookie:
-      key: string
-      value: ByteArray
-   # MC: ClientboundTransferPacket
-   packet_common_transfer:
-      host: string
-      port: varint
-   # MC: ServerboundCookieResponsePacket
-   packet_common_cookie_response:
-      key: string
-      value: ByteArray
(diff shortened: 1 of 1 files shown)


Slauh:
interesting. does It give vaild data?

rom1504:
I think so, you can see in this PR

Maks-gaming:
@rom1504 it seems now it's a useless PR since wiki.vg is closed

rom1504:
@Maks-gaming you mean it became an essential PR since it may allow to keep the comments of wiki.vg now that it is closed ?

Maks-gaming:
> @Maks-gaming you mean it became an essential PR since it may allow to keep the comments of wiki.vg now that it is closed ?

ah, I see what you mean - I thought it was useless because the site is closed, but you’re right: it actually makes sense as a way to preserve those comments, nevermind

PR #988, opened 2025-02-26
Bedrock ItemStackRequest missing some "times_crafted" fields

EntifiedOptics:
In the Bedrock protocol the ItemStackRequest type is missing two `times_crafted` fields in `craft_grindstone_request` and `craft_loom_request` ([Reference](https://github.com/Sandertv/gophertunnel/blob/bfe0892926356f3aeb922df94a2e909804ea007e/minecraft/protocol/item_stack.go))

Unrelated question, is there any way to change all effected versions any other way rather than changing each file one by one? Apologies if I'm asking a silly question, I am not too familiar with github features.

### DIFF
--- a/data/bedrock/1.21.50/protocol.json
+++ b/data/bedrock/1.21.50/protocol.json
@@ -2918,6 +2918,10 @@
                                 "type": "varint"
                               },
                               {
+                                "name": "times_crafted",
+                                "type": "u8"
+                              },
+                              {
                                 "name": "cost",
                                 "type": "varint"
                               }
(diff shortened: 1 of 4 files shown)


extremeheat:
> Unrelated question, is there any way to change all effected versions any other way rather than changing each file one by one? Apologies if I'm asking a silly question, I am not too familiar with github features.

Yes, if you're using VS Code, you can select the relevant block of old text, copy it, and use the Find All feature (Ctrl+Shift+F) to find and replace the relevant instances for all the relevant versions with the desired new text.

extremeheat (approved):

extremeheat:
Thanks!

PR #999, opened 2025-04-03
Bedrock protocol missing debug_info packet

EntifiedOptics:
All bedrock protocols past 1.16 are missing the `debug_info` packet (From reference [here](https://github.com/Sandertv/gophertunnel/blob/bfe0892926356f3aeb922df94a2e909804ea007e/minecraft/protocol/packet/debug_info.go)).


extremeheat (approved):

extremeheat:
Thanks

PR #1000, opened 2025-04-05
Bedrock protocol added missing code_builder packet

EntifiedOptics:
The `code_builder` packet is missing in all post-1.16 bedrock protocols. ([Reference](https://github.com/Sandertv/gophertunnel/blob/bfe0892926356f3aeb922df94a2e909804ea007e/minecraft/protocol/packet/code_builder.go))


extremeheat:
Thanks

PR #1021, opened 2025-06-17
Add bedrock 818 (1.21.90) protocol data

CreeperG16:


CreeperG16:
There seemingly also have been some changes to the login packet, specifically within the token chain. They aren't documented in the official docs, but cloudburst have implemented it somewhat

CreeperG16:
Looking at what gophertunnel is doing, auth can probably use the legacy auth for now, with a few changes in the JSON structure (just put the legacy chains in the Certificate key)
I started work last night on the changes to bedrock-protocol, but I won't have time until later this evening to continue

extremeheat:
Feel free to make a PR in bedrock-protocol if it's needed

CreeperG16:
Will do once I get home

extremeheat:
Thanks for the work!

PR #1032, opened 2025-07-04
Bedrock v1.21.93

CreeperG16:
This hotfix bumps the protocol version but there don't seem to be any documented changes in any first- or third party projects


CreeperG16:
ah, well I just ran incrementVersion.js and left it at that; good to know, I'll keep it in mind for next time. thanks

PR #1070, opened 2025-08-27
🎈 Pc1.21.8

extremeheat:
(Not for merge, there are no protocol changes in 1.21.8, this is just for testing workflows)


rom1504:
Would be good to actually have a merge able 1.21.8

extremeheat:
1.21.7 and 1.21.8 are same protocol, I checked the code diff and only change is data so codewise current minecraft-data should already support 1.21.7 and 1.21.8

data/pc/1.21.8/blocks.json
```diff
@@ -2703,7 +2703,9 @@
         "num_values": 2
       }
```
extremeheat, line 2704: Looks like only change in .8 is block drops, the entities.json changed because the entity metadata extractor was not run in other PR

PR #1081, opened 2025-09-21
bedrock protocol: use ZigZag64 instead of VarInt64 in packet_player_location

Snowbelldog:
fix(protocol): use ZigZag64 instead of VarInt64 for entity_unique_id at packet_player_location

The entity_unique_id field in packet_player_location was incorrectly
deserialized as VarInt64. It is now handled as ZigZag64, fixing mismatched
IDs when comparing with player_list.


extremeheat:
Thanks, does this apply to any other versions?

Snowbelldog:
I checked the CloudburstMC Java implementation used in GeyserMC. There is only a deserializer for this starting from protocl version v800 [(link)](https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/codec/v800/serializer/PlayerLocationSerializer_v800.java). The implementation at first looks like a regular VarInt, but on closer inspection it’s actually a ZigZag64. Therefore, in previous versions this was also incorrect.

Snowbelldog:
I also noticed that in packet_update_trade there seems to be a similar issue, including the villager_unique_id field. However, I think it would be out of scope here.

extremeheat (approved):

extremeheat:
Thanks

PR #1083, opened 2025-09-22
fix packet_chat_message checksum type in PC V1.21.8

kang-indie-hacker:
issue description: it leads to RangeError when chat 
details can be seen in the follow url
https://github.com/PrismarineJS/mineflayer/issues/3724


DereC4:
> Are there any other affected versions?

Only happening to me on 1.21.8

PR #1084, opened 2025-09-23
Steve skin version didn't properly update after 1.21.70

thejfkvis:
The Steve Skin data was last updated in v1.21.70, Someone had forgot to change the Steve Skin Data version after v1.21.70 to include the more updated one


extremeheat:
Thanks

Issue #981, opened 2025-01-29
There is no identifier for the "Projectile Power" packet in version 1.20.5


Titlehhhh:
https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol?oldid=2773283#Projectile_Power

As you can see the packet is there, but...
https://github.com/PrismarineJS/minecraft-data/blob/10749e0a06563f0cd563327e85c003b145923648/data/pc/1.20.5/protocol.json#L7322

but
https://github.com/PrismarineJS/minecraft-data/blob/10749e0a06563f0cd563327e85c003b145923648/data/pc/1.20.5/protocol.json#L7465

Titlehhhh:
in 1.21.3 there is also an error in set_projectile

PR #1088, opened 2025-09-29
fix(protocol): add missing packet ids

Titlehhhh:
I'm writing a C# packet generator based on minecraft-data. 
During validation I found that some packets were described in the protocol schema 
but had no IDs assigned in the `packet` mapping. Because of this my generator broke.  

This PR fixes the [issue ](https://github.com/PrismarineJS/minecraft-data/issues/981) by adding the missing packet IDs.  
I already created an issue about this a while ago, but it was ignored, 
so I'm opening a PR directly.

### DIFF
--- a/data/pc/1.20.5/protocol.json
+++ b/data/pc/1.20.5/protocol.json
@@ -7237,7 +7237,8 @@
                     "0x75": "entity_update_attributes",
                     "0x76": "entity_effect",
                     "0x77": "declare_recipes",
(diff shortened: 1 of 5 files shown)


extremeheat:
CI is failing

Titlehhhh:
Should I add yaml?

extremeheat:
Yes, the json is generated from the yaml

PR #1114, opened 2025-11-29
Fix packet_explosion partial packet issue

Rohan2460:
`packet_explosion` was reading 32 bit vec3 for playerKnockback instead of 64 bit vec3. 
This was causing issues with knockback calculation. Mineflayer bot  in survival mode when affected by explosion knockback results in a crash.  
https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Protocol?oldid=3258009#Explosion

Versions updated
- 1.21.8
- 1.21.7
- 1.21.6
- 1.21.5
- 1.21.4

Fixes https://github.com/PrismarineJS/mineflayer/issues/3753

### DIFF
--- a/data/pc/1.21.4/proto.yml
+++ b/data/pc/1.21.4/proto.yml
@@ -1736,7 +1736,7 @@
       x: f64
       y: f64
       z: f64
-      playerKnockback?: vec3f
+      playerKnockback?: vec3f64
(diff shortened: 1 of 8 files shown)


extremeheat:
Thanks

PR #1155, opened 2026-03-07
Fix attribute_modifiers structure in protocol.json for 1.21.5-1.21.11

reallyoldfogie:
The attribute_modifiers component was incorrectly structured as a container with "attributes" and "display" as sibling top-level fields. Based on an analysis of AttributeModifiersComponent in the affected versions, the correct structure is an array of entries where each entry contains typeId, name, value, operation, slot, and display fields inline (except 1.21.5, which doesn't have a display).

Affected versions: 1.21.5, 1.21.6, 1.21.8, 1.21.9, 1.21.11


data/pc/1.21.5/proto.yml
```diff
@@ -185,28 +185,26 @@
...
-               - off_hand
-               - hand
-               - feet
-               - legs
-               - chest
-               - head
-               - armor
-               - body
-               - saddle
-            showTooltip: bool
```
extremeheat: Erroneously removed?

extremeheat:
Are you sure this is necessary? Seems to me you are only unwrapping a field which should not affect serialization/deserialization beyond an extra container. There are changes between versions also, see comment

extremeheat:
I see, so main change here (aside from cosmetic unwrapping) is the correction to 1.21.5+ in that `showInTooltip`/`showTooltip` field was removed from `attribute_modifiers` mode of SlotComponent.
1.21.6+ seem to already remove it but 1.21.5 retained it. Thanks for investigation, LGTM.

PR #1170, opened 2026-04-01
blocks.json minecraft bedrock 26.10

M9CHKO:
updated blocks.json for the minecraft bedrock version 26.10


extremeheat:
Would be good to automate this, we don't have a means to test or validate this at the moment

PR #1180, opened 2026-05-06
fix(bedrock/proto): mismatched field type in GatheringJoinInfo

hax0r31337:
scenario_id should be string instead of uuid
https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/codec/v944/serializer/StartGameSerializer_v944.java#L26

### DIFF
--- a/data/bedrock/26.10/protocol.json
+++ b/data/bedrock/26.10/protocol.json
@@ -5202,12 +5202,12 @@
           "type": "string"
         },
         {
-          "name": "unknown_uuid_1",
+          "name": "target_id",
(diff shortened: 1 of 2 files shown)


extremeheat:
Can you apply to previous versions also?

hax0r31337:
> Can you apply to previous versions also?

This struct is just got added in 1.26.10 (protocol v944)

extremeheat (approved):

extremeheat:
Thanks

PR #1181, opened 2026-05-09
Add bedrock 1.26.20 protocol data

extremeheat:
Rename old 26.10 to 1.26.10 for consistency with vanilla internals


data/bedrock/latest/types.yml
```diff
@@ -819,15 +833,6 @@ Recipes: []varint
          assume_symmetry: bool
          unlocking_requirement: RecipeUnlockingRequirement
          network_id: varint
-      if furnace:
```
extremeheat, line 822: undo

PR #1187, opened 2026-05-13
Fix checksum error boundary in 1.21.9 and 1.21.11.

### LINKED ISSUE #1488: Error boundary for checksum in 1.21.11
- mineflayer latest
- minecraft-version 1.21.11

## problem
 In 1.21.11, the following error will be thrown after using `bot.chat` several times.
```txt
RangeError [ERR_OUT_OF_RANGE]: Write error for undefined : The value of "value" is out of range. It must be >= 0 and <= 255. Received -9
    at writeU_Int8 (node:internal/buffer:740:11)
    at Buffer.writeUInt8 (node:internal/buffer:750:10)
    at Object.writer [as u8] (D:\Code\mc\mineflayer\test\node_modules\protodef\src\datatypes\numeric.js:97:25)
    at Object.packet_chat_message (eval at compile (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:262:12), <anonymous>:3208:24)
    at eval (eval at compile (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:262:12), <anonymous>:3835:64)
    at packet (eval at compile (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:262:12), <anonymous>:3895:9)
    at CompiledProtodef.write (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:76:12)
    at e.message (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:102:25)
    at tryCatch (D:\Code\mc\mineflayer\test\node_modules\protodef\src\utils.js:50:16)
    at CompiledProtodef.createPacketBuffer (D:\Code\mc\mineflayer\test\node_modules\protodef\src\compiler.js:102:5) {
  code: 'ERR_OUT_OF_RANGE',
  field: 'play.toServer'
}
```

The `chat_message` packet is:
```txt
{
  message: '5',
  timestamp: 1778598229848n,
  salt: 1n,
  signature: <Buffer 2a 03 ba e0 e5 ef 6a 3e 07 6d 1c 8a 66 5e 6d 95 7e 08 e8 3d da fb 60 73 f6 9c fd 8a e6 d2 5c 51 22 8a 1f 0c 09 33 74 85 86 2c ae 3b 94 db 67 0b 36 cf ... 206 more bytes>,
  offset: 1,
  checksum: -9,
  acknowledged: <Buffer 00 00 0c>
}
```

#1442 mention:
> The Minecraft 1.21.5+ protocol expects checksums to be signed bytes (range -128 to 127) rather than unsigned bytes (range 0-255)

I think the checksum boundary need to update to -128 to 127 in 1.21.5+, but I don't know how to fix it...

TAOtxi:
fix PrismarineJS/node-minecraft-protocol#1488

### DIFF
--- a/data/pc/1.21.11/protocol.json
+++ b/data/pc/1.21.11/protocol.json
@@ -10466,7 +10466,7 @@
             },
             {
               "name": "checksum",
-              "type": "u8"
+              "type": "i8"
             }
           ]
         ],
--- a/data/pc/1.21.9/proto.yml
+++ b/data/pc/1.21.9/proto.yml
@@ -3324,7 +3324,7 @@
             "count": 3
          }
       ]
-      checksum: u8
+      checksum: i8
    # MC: ServerboundChatSessionUpdatePacket
    packet_chat_session_update:
       sessionUUID: UUID
--- a/data/pc/1.21.9/protocol.json
+++ b/data/pc/1.21.9/protocol.json
@@ -10176,7 +10176,7 @@
             },
             {
               "name": "checksum",
-              "type": "u8"
+              "type": "i8"
             }
           ]
         ],
--- a/data/pc/latest/proto.yml
+++ b/data/pc/latest/proto.yml
@@ -3401,7 +3401,7 @@
             "count": 3
          }
       ]
-      checksum: u8
+      checksum: i8
    # MC: ServerboundChatSessionUpdatePacket
    packet_chat_session_update:
       sessionUUID: UUID


extremeheat:
Why is the checksum an i8 and not u8? That almost certainly seems wrong

Come to think of it https://github.com/PrismarineJS/node-minecraft-protocol/pull/1442 doesn't seem to make sense either. Is there some inconsistency?

TAOtxi:
@extremeheat [wiki](https://minecraft.wiki/w/Java_Edition_protocol/Packets#Chat_Message) mentioned checksum is sign byte.

extremeheat:
Java doesn't have unsigned types. This is not relevant to a checksum which are almost always treated as unsigned. It doesn't make sense to have a signed checksum

extremeheat:
A 32 bit checksum will use all 32 bits, there is no "sign bit" left over

extremeheat:
Looks like the root issue is that there are different checksum fields with a mismatch of i8/u8... this should be fixed and nmp PR can be reverted

TAOtxi:
But https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.21.8/protocol.json#L9045 is also i8, and this commit fix the issue I mentioned.😐

extremeheat:
Yes, it's incorrectly i8 in some places and u8 in others. Being fixed in #1188 and https://github.com/PrismarineJS/node-minecraft-protocol/pull/1489

PR #1191, opened 2026-05-17
Bedrock_1.20.0+, biomes fix.

nootbook011:
Fixed biomes data mismatch with bedrock versions 1.20.0 - 1.21.60.


extremeheat:
Thanks

extremeheat:
We should automate this data generation at some point

PR #1192, opened 2026-05-19
update bedrock data 1.26.0-1.26.30

mc-zuri:


extremeheat:
As mentioned prior, we really need to either automate this or have some means to test

extremeheat:
We may in future have to remove some data that can't be obtained through an automated step, but we can merge the data for now

PR #1199, opened 2026-06-17
Support for 1.26.30

MooMoo2026-26:


extremeheat:
Thanks for the work on this, though seems this is missing a bit of changes on protocol side

extremeheat:
Being done in #1202

PR #1209, opened 2026-07-22
Fix complete 1.21.4 attribute registry

iiroak:
## Summary
- restore the complete 32-entry Java 1.21.4 attribute registry
- correct movement speed to protocol ID 21 and step height to ID 27
- regenerate the JSON protocol definition
- add a regression test for the affected IDs

The incomplete table shifted attributes after ID 8, causing movement speed packets to decode as step height.

## Validation
- `npm run build -- pc 1.21.4`
- `npx mocha test/audit_attributes.js test/protocolSync.js --exit`
- `npm run lint`


extremeheat:
Does this only affect 1.21.4 ?

PR #1215, opened 2026-08-09
Fix player_list + add new DisconnectFailReasons for v1.26.40

thejfkvis:
Last PR for v1.26.40 removed player_list data that still exists on this version, plus I updated the DisconnectFailReasons.

Tests will fail because the bot doesn't have a DeviceOS field and this bug only exists on 1.26.40.8 of BDS, as they blindly accepted the DeviceOS field at the time without any validation until version 1.26.43.1 of BDS.

### DIFF
--- a/data/bedrock/1.26.40/protocol.json
+++ b/data/bedrock/1.26.40/protocol.json
@@ -2165,7 +2165,7 @@
         },
         {
           "name": "skin_color",
-          "type": "i32"
+          "type": "li32"
         },
         {
           "name": "personal_pieces",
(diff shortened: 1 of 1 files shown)


extremeheat:
The CI isn't failing because of a device ID, it's failing because the serializer schema is wrong:

```
      at Object.packet_player_list 
```

The server is likely not sending the data, and since this is enclosed in an array, it's not just extra data at end of the packet

extremeheat:
Thanks, being done in https://github.com/PrismarineJS/minecraft-data/pull/1217

PR #1230, opened 2026-08-16
pc: fix custom_click_action NBT to use a VarInt length prefix

### LINKED ISSUE #1222: PC 1.21.8: custom_click_action NBT is missing its VarInt length prefix
### Version

Minecraft Java 1.21.8, protocol 772.

### Problem

The current `packet_common_custom_click_action` schema is:

 ```yml
packet_common_custom_click_action:
    id: string
    nbt?: anonymousNbt
 ```
This encodes the NBT field as an optional value with a boolean prefix.

However, the actual wire format used by the vanilla client is:

resource location + VarInt payload length + unnamed NBT payload

For example, for a 27-byte NBT payload the correct packet tail starts with:
1B 0A 0B ...

Where 1B is the VarInt payload length.

The current schema produces:
01 0A 0B ...

BungeeCord interprets 01 as the payload length and gives only one byte to
the NBT decoder, resulting in:

DecoderException: RuntimeException: Exception reading tag

Writing the anonymous NBT without any prefix produces:
0A 0B ...

and is also invalid because 0A is interpreted as the payload length.

### Verified workaround

Changing the field to a VarInt-length-prefixed buffer:

```yml
packet_common_custom_click_action:
    id: string
    nbt: ["buffer", { "countType": "varint" }]
```
and serializing the unnamed NBT into that buffer produced the same bytes as the
vanilla client. The dialog action was successfully accepted by the server.

The incorrect definition is here:

https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.21.8/proto.yml#L1330-L1333

BungeeCord also reads this value through a length-prefixed NBT decoder:

https://github.com/SpigotMC/BungeeCord/blob/master/protocol/src/main/java/net/md_5/bungee/protocol/packet/CustomClickAction.java

AnonymoDGH:
Fixes #1222

## Problem
`packet_common_custom_click_action` declared its NBT field as:

```yml
packet_common_custom_click_action:
    id: string
    nbt?: anonymousNbt
```

which encodes it as an *optional* value with a boolean prefix. The actual wire format used by the vanilla client (and expected by BungeeCord) is:

```
resource location + VarInt payload length + unnamed NBT payload
```

BungeeCord reads it via `readLengthPrefixed(...)` (`CustomClickAction.java`), i.e. a VarInt length followed by the NBT bytes — there is no optional sentinel. With the old schema the client wrote `01 0A 0B ...` (BungeeCord read `01` as the payload length and the NBT decoder failed with `Exception reading tag`), and writing the NBT without any prefix (`0A 0B ...`) is also invalid.

## Fix
Changed the field to a VarInt-length-prefixed buffer in every affected version — the exact representation verified in the issue to produce byte-identical output to the vanilla client:

```yml
packet_common_custom_click_action:
    id: string
    nbt: ["buffer", { "countType": "varint" }]
```

Applied to `data/pc/1.21.6`, `1.21.8`, `1.21.9`, `1.21.11` and `latest` (all of which carried the same incorrect definition), and regenerated the corresponding `protocol.json` files (`1.21.6`, `1.21.8`, `1.21.9`, `1.21.11`, `26.1`) with `npm run build`.

## Verified
- Cross-checked against BungeeCord's `CustomClickAction` (`readLengthPrefixed`/`writeLengthPrefixed` → VarInt length + NBT, always present).
- `protocolSync` test passes: all proto.yml files are valid and in sync with their protocol.json.
- The regenerated `protocol.json` diffs contain only the `nbt` field change.

### DIFF
--- a/data/pc/1.21.11/proto.yml
+++ b/data/pc/1.21.11/proto.yml
@@ -1638,7 +1638,7 @@
    # MC: ServerboundCustomClickActionPacket
    packet_common_custom_click_action:
       id: string
-      nbt?: anonymousNbt
+      nbt: ["buffer", { "countType": "varint" }]
 
 ^handshaking.toClient.types:
    packet:
--- a/data/pc/1.21.11/protocol.json
+++ b/data/pc/1.21.11/protocol.json
@@ -4937,8 +4937,10 @@
         {
           "name": "nbt",
           "type": [
-            "option",
-            "anonymousNbt"
+            "buffer",
+            {
+              "countType": "varint"
+            }
           ]
         }
       ]
--- a/data/pc/1.21.6/proto.yml
+++ b/data/pc/1.21.6/proto.yml
@@ -1330,7 +1330,7 @@
    # MC: ServerboundCustomClickActionPacket
    packet_common_custom_click_action:
       id: string
-      nbt?: anonymousNbt
+      nbt: ["buffer", { "countType": "varint" }]
 
 ^handshaking.toClient.types:
    packet:
--- a/data/pc/1.21.6/protocol.json
+++ b/data/pc/1.21.6/protocol.json
@@ -3611,8 +3611,10 @@
         {
           "name": "nbt",
           "type": [
-            "option",
-            "anonymousNbt"
+            "buffer",
+            {
+              "countType": "varint"
+            }
           ]
         }
       ]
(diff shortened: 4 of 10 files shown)


AnonymoDGH:
Note on CI: the `build` check (bedrock-ci.yml) fails with `Cannot find package '@napi-rs/snappy-wasm32-wasi'` while loading the **separate** `bedrock-protocol` repo's own modules. This is an environmental dependency issue in that repo's test setup and is unrelated to this change, which only touches `data/pc/*` (PC protocol data) that bedrock-protocol tests don't consume. The authoritative validation for this change — `build (24)` (ci.yml → tools/js protocolSync + schema tests) — passes.

PR #1194, opened 2026-06-12
schemas: add registry tags schema and audit

VasilisDragon:
## Problem

This PR was originally based on #1080's 1.21.8 four-registry output. The replacement generator work uses full resource locations and can emit any keyed registry, so the schema and repository checks need to cover the final representation.

## Fix

- validate full registry, tag and member resource locations while preserving empty tags;
- reject duplicate members and unresolved `#tag` references;
- add a PC-only audit for canonical ordering, `dataPaths` targets, orphan files and mapped member references;
- keep additional registry types open while requiring block and item registries in tag-bearing PC data.

## Test

- focused positive, negative and corruption controls pass;
- the full `tools/js` suite passes;
- the schema compiles to the generated `Tags` type;
- duplicate validation adds about 9 ms across the three tested tag artifacts.

## Scope

Schema and audit only. No generated tags data or `dataPaths` changes are included here.

## Refs

- PrismarineJS/minecraft-data-generator#56
- #1080
- #463
- #464
- PrismarineJS/node-minecraft-data#457

PR #1232, opened 2026-08-16
fix: add valid pickaxes to incorrect_for_wooden_tool material map (wrong dig times)

### LINKED ISSUE #3921: wrong dig time when block material doesn't have tool IDs mapped in registry
hey , found a bug with dig time calculation for certain blocks.
some blocks have a material (like incorrect_for_wooden_tool) where canHarvest works correctly , but the speed lookup fails silently and falls back to 1. result is the bot takes 4-5x longer than it should to dig the block.
the issue is in how prismarine-block calculates speed:

takes block.material
looks up registry.materials[block.material]
checks if the held tool ID exists in that material
if found → uses tool speed , if not → fallback to speed 1

for some blocks the material exists in the registry but doesnt have the actual tool IDs mapped , so every tool that can harvest the block still gets speed 1.
debug output:
hardness: 3
material: incorrect_for_wooden_tool
canHarvestHeld: true
heldItem: iron_pickaxe (type 933)
botDigTime: 4550ms
toolComparison shows all valid pickaxes returning the same wrong value , confirming the speed lookup fails for all of them.
with speed 1 (bugged):
hardness=3 , speed=1 , ticks≈90 → ~4550ms
with speed 6 (correct):
hardness=3 , speed=6 , ticks=15 → ~750ms
so canHarvest is correct , only the speed lookup is broken.

minecraft version: [1.21.11]
mineflayer version: [4.37.1]

AnonymoDGH:
## Problem

Reported in PrismarineJS/mineflayer#3921: on 1.21.11, `block.digTime()` returns wildly wrong values for blocks whose material is `incorrect_for_wooden_tool` (iron ore, copper ore, obsidian, ancient debris, raw metal blocks, etc.) when holding a valid pickaxe.

Reproduction with prismarine-block + minecraft-data 3.113.2 (1.21.1, iron_ore, hardness 3):

```
wooden_pickaxe:   canHarvest=undefined digTime=7500ms
stone_pickaxe:    canHarvest=true      digTime=4550ms   <- wrong
iron_pickaxe:     canHarvest=true      digTime=4550ms   <- wrong (vanilla: 750ms)
diamond_pickaxe:  canHarvest=true      digTime=4550ms   <- wrong
netherite_pickaxe: canHarvest=true     digTime=4550ms   <- wrong
```

## Root cause

`prismarine-block`'s `digTime` looks the held item up in `registry.materials[block.material]`. The `incorrect_for_wooden_tool` map only contains the four wooden tools (the *incorrect* ones), so every valid pickaxe misses the lookup and silently falls back to a block-breaking speed of 1 instead of its real speed (stone 4, iron 6, diamond 8, netherite 9).

## Fix

For every version that has this material (1.20.5, 1.21.1, 1.21.3, 1.21.4, 1.21.5, 1.21.6, 1.21.8, 1.21.9, 1.21.11, 26.1), add the non-wooden pickaxes from `mineable/pickaxe` into `incorrect_for_wooden_tool` at their correct speeds. The existing wooden-tool entries are preserved untouched (including the 1.21.4 `wooden=1.0` values introduced by #1033). On 1.21.9+ this also adds `copper_pickaxe` at speed 1, matching its `mineable/pickaxe` entry.

Example (1.21.11):
```diff
   "incorrect_for_wooden_tool": {
     "912": 2.0,
     "913": 2.0,
     "914": 2.0,
-    "915": 2.0
+    "915": 2.0,
+    "918": 1.0,
+    "923": 4.0,
+    "928": 12.0,
+    "933": 6.0,
+    "938": 8.0,
+    "943": 9.0
   },
```

## Verification

Schema: all 10 changed files pass `materials_schema.json` validation (ajv).

End-to-end via prismarine-block `digTime` after this change (1.21.1 and 1.21.11, iron_ore):

```
wooden_pickaxe:   canHarvest=undefined digTime=7500ms   (unchanged, no drops)
stone_pickaxe:    canHarvest=true      digTime=1150ms
golden_pickaxe:   canHarvest=undefined digTime=1250ms   (fast but no drops, as in vanilla)
iron_pickaxe:     canHarvest=true      digTime=750ms    <- matches vanilla
diamond_pickaxe:  canHarvest=true      digTime=600ms
netherite_pickaxe: canHarvest=true     digTime=500ms
```

`canHarvest` (drop gating via `harvestTools`) is unaffected; this only corrects the mining speed lookup.

Fixes PrismarineJS/mineflayer#3921 (data side).


AnonymoDGH:
Note on CI: the Node 22 `build` job fails in the **bedrock-protocol** test step with `Cannot find package '@napi-rs/snappy-wasm32-wasi'` — a dependency-resolution error at module load, before any data file is read. This is unrelated to this data-only change and is pre-existing: the same failure appears on #1231 and #1230 (and the Node 22 job has been red across recent PRs).

The `build (24)` job, which runs the full data test suite including `materials_schema.json` validation, **passes** here. All 10 changed `materials.json` files also validate against the schema locally (ajv).

extremeheat:
Note - optimal thing in future is to remove the legacy material.json handling and replace it with tag system that modern versions use

#463 , https://github.com/PrismarineJS/minecraft-data/pull/1194

PR #1234, opened 2026-08-16
pc: fix warped_fungus_on_a_stick stack size to 1 (1.16.1-1.18)

### LINKED ISSUE #569: Wrong stack size for Warped Fungus on a Stick
Warped Fungus on a Stick's stack size should be 1, not 64.

https://github.com/PrismarineJS/minecraft-data/blob/d5771849dd28d110dcb587e4c3a5913d9888049f/data/pc/1.18/items.json#L4013-L4023

AnonymoDGH:
Fixes #569

The Warped Fungus on a Stick has a max stack size of **1** in vanilla Minecraft (it has durability, maxDurability 100). The data incorrectly listed it with stackSize 64 in versions 1.16.1, 1.16.2, 1.17 and 1.18.

Verified: item entry in each version's items.json now has `"stackSize": 1`. JSON is valid; schema validation passes.

### DIFF
--- a/data/pc/1.16.1/items.json
+++ b/data/pc/1.16.1/items.json
@@ -5675,7 +5675,7 @@
     "id": 842,
     "displayName": "Warped Fungus on a Stick",
     "name": "warped_fungus_on_a_stick",
-    "stackSize": 64,
+    "stackSize": 1,
     "maxDurability": 100,
     "enchantCategories": [
       "breakable",
--- a/data/pc/1.16.2/items.json
+++ b/data/pc/1.16.2/items.json
@@ -5681,7 +5681,7 @@
     "id": 843,
     "displayName": "Warped Fungus on a Stick",
     "name": "warped_fungus_on_a_stick",
-    "stackSize": 64,
+    "stackSize": 1,
     "maxDurability": 100,
     "enchantCategories": [
       "breakable",
--- a/data/pc/1.17/items.json
+++ b/data/pc/1.17/items.json
@@ -4014,7 +4014,7 @@
     "id": 668,
     "displayName": "Warped Fungus on a Stick",
     "name": "warped_fungus_on_a_stick",
-    "stackSize": 64,
+    "stackSize": 1,
     "maxDurability": 100,
     "enchantCategories": [
       "breakable",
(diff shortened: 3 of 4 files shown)


extremeheat:
Note that this data is typically generated from minecraft-data-generator server (https://github.com/PrismarineJS/minecraft-data-generator). Because this is an old version data we can probably merge, but changes made to newer versions will automatically revert to the previous data state without updating the data generator

AnonymoDGH:
Thanks for the note! This fix only touches 1.16.1-1.18 (pre-1.19 versions that the generator no longer refreshes), so it won't conflict with generated data. I'll keep this in mind for any newer-version changes.

AnonymoDGH:
I'm going to improve the mod from https://github.com/PrismarineJS/minecraft-data-generator; I want to see if I can create a C program that extracts data from the Minecraft JAR without having to launch the game.

extremeheat:
Most of the data is not stored in some string inside the jar, it's programmatically stored inside generator functions and includes runtime components. So in order to get at much of the data, you need to execute the code which is what the data gen does. It's not a simple JSON parser

AnonymoDGH:
I opened a PR on minecraft-data-generator (PrismarineJS/minecraft-data-generator#79) that adds a DataOverrides layer in the shared generator pipeline, so regenerated versions keep these corrections (stackSize, copper_ore, water/lava diggable, effect ids) instead of reverting to the previous data state.

AnonymoDGH:
Good news on the revert concern — I checked the current `minecraft-data-generator` `main` output and it **already generates `stackSize: 1`** for `warped_fungus_on_a_stick`:

| Version (generator main) | warped_fungus_on_a_stick stackSize |
|---|---|
| 1.16 | 1 |
| 1.17 | 1 |
| 1.18 | 1 |

(Verified from the generator's own CI artifacts on `main`.)

So the generator is already correct here — the master data files are just stale relative to what the generator now produces. Merging this data fix will **not** be reverted by a future regeneration, since regenerating yields the same `stackSize: 1` value this PR sets. The data and the generator are already in agreement; this PR just brings the stored data up to match.

PR #1241, opened 2026-08-16
test: enforce metadataKeys presence in pc entities 1.19.4+

### LINKED ISSUE #1058: Validate presence of entity medata keys
missing in https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.21.6/entities.json

AnonymoDGH:
Fixes #1058

Adds an automated audit test that enforces the presence of the `metadataKeys` key in `entities.json` for all PC versions from 1.19.4 onward, as suggested in the [mineflayer PR comment](https://github.com/PrismarineJS/mineflayer/pull/3713#issuecomment-3193880222).

Verified: 127 tests pass; all current PC entities 1.19.4+ already have metadataKeys.


extremeheat:
Good idea, but we should also probably add this generation step to the current new version automation

PR #588, opened 2022-07-18
Bedrock 1.19.10 Initial Commit

stevarino:
# Works

Followed the following PRs for data:

 - https://github.com/Sandertv/gophertunnel/pull/136/files 
 - https://github.com/CloudburstMC/Protocol/pull/162/files

Interesting new packets:

```
update_adventure_settings {
  no_pvm: false,
  no_mvp: false,
  imutable_world: false,
  show_name_tags: false,  # should this be true?
  auto_jump: true  # also this seems an odd default?
}

update_abilities
{
  "entity_id": [  # unsure why this is an array
    -3,
    1
  ], 
  "permission_level": "member",
  "command_permission": "normal",
  "ability_layers": [
    {
      "layer_type": "base",
      "abilities": {
        "_value": 262143,
        "build": true,
        "mine": true,
        "doors_and_switches": true,
        "open_containers": true,
        "attack_players": true,
        "attack_mobs": true,
        "operator_commands": true,
        "teleport": true,
        "invulnerable": true,
        "flying": true,
        "may_fly": true,
        "instant_build": true,
        "lightning": true,
        "fly_speed": true,
        "walk_speed": true,
        "muted": true,
        "world_builder": true,
        "no_clip": true,
        "ability_count": false
      },
      "values": {
        "_value": 63,
        "build": true,
        "mine": true,
        "doors_and_switches": true,
        "open_containers": true,
        "attack_players": true,
        "attack_mobs": true,
        "operator_commands": false,
        "teleport": false,
        "invulnerable": false,
        "flying": false,
        "may_fly": false,
        "instant_build": false,
        "lightning": false,
        "fly_speed": false,
        "walk_speed": false,
        "muted": false,
        "world_builder": false,
        "no_clip": false,
        "ability_count": false
      },
      "fly_speed": 0.05000000074505806,
      "walk_speed": 0.10000000149011612
    }
  ]
}
```

PR #1248, opened 2026-08-16
bedrock: fix start_game permission_level type (zigzag32) for 1.19.10+

### LINKED ISSUE #724: start_game seed and biome are incorrect?
Is it expected that the seed and biome information will be wrong in packet_start_game when connecting to a realm or is there a parsing error?  If it is expected, is the correct information available somewhere?  The position information that precedes it seems correct aside from the off-by-one Y value which I assume is just the difference between head and foot or player and camera location.  The dimension that comes after is correct but the permission level seems wrong as well. Other than world name and the engine version which look correct, I can't verify much else in the packet.

I'm attempting to look at the parsing to figure it out myself but I'm seeing conflicting information about the packet structure and a lack of specification for the encoding of some of the types in the packet.

AnonymoDGH:
Fixes #724

As confirmed by @extremeheat, the `player_permissions` field of `packet_start_game` was incorrectly changed from `zigzag32` to `PermissionLevel` (u8) in PR #588 (starting 1.19.10), breaking the permission level read on all later versions.

Verified against [pmmp LevelSettings](https://github.com/pmmp/BedrockProtocol/blob/master/src/types/LevelSettings.php): `defaultPlayerPermission` uses `VarInt::readSignedInt` (zigzag32). Versions up to 1.19.1 already used `zigzag32`.

This reverts `permission_level: PermissionLevel` to `permission_level: zigzag32` in start_game for all 39 affected versions (1.19.10 → latest). Other packets using PermissionLevel are unchanged.

Verified: protocol regenerated for all versions; only the one field line changed per version.

### DIFF
--- a/data/bedrock/1.19.10/proto.yml
+++ b/data/bedrock/1.19.10/proto.yml
@@ -311,7 +311,7 @@ packet_start_game:
    map_enabled: bool
    # The permission level of the player. It is a value from 0-3, with 0 being visitor,
    # 1 being member, 2 being operator and 3 being custom.
-   permission_level: PermissionLevel
+   permission_level: zigzag32
    # The radius around the player in which chunks are ticked. Most servers set this value
    # to a fixed number, as it does not necessarily affect anything client-side.
    server_chunk_tick_range: li32
--- a/data/bedrock/1.19.10/protocol.json
+++ b/data/bedrock/1.19.10/protocol.json
@@ -4845,7 +4845,7 @@
         },
         {
           "name": "permission_level",
-          "type": "PermissionLevel"
+          "type": "zigzag32"
         },
         {
           "name": "server_chunk_tick_range",
--- a/data/bedrock/1.19.20/proto.yml
+++ b/data/bedrock/1.19.20/proto.yml
@@ -311,7 +311,7 @@ packet_start_game:
    map_enabled: bool
    # The permission level of the player. It is a value from 0-3, with 0 being visitor,
    # 1 being member, 2 being operator and 3 being custom.
-   permission_level: PermissionLevel
+   permission_level: zigzag32
    # The radius around the player in which chunks are ticked. Most servers set this value
    # to a fixed number, as it does not necessarily affect anything client-side.
    server_chunk_tick_range: li32
--- a/data/bedrock/1.19.20/protocol.json
+++ b/data/bedrock/1.19.20/protocol.json
@@ -4885,7 +4885,7 @@
         },
         {
           "name": "permission_level",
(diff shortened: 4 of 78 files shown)


extremeheat:
The data type seems correct as is, it's a bounded byte but encoded as a "varint8". Since we only support 32 bit varints, the correct data type for now is closer to u8 although it can also technically overflow on illegal values >128

Copilot:
---
**Mojang’s official reference does not support this PR’s blanket `zigzag32` change.**

The current official Bedrock protocol documentation defines `StartGamePacket → LevelSettings → Player Permissions` as:

```text
varint8
```

at field index 28, with values:

- `0` Visitor
- `1` Member
- `2` Operator
- `3` Custom

See Mojang’s [`StartGamePacket` reference](https://github.com/Mojang/bedrock-protocol-docs/blob/7330880ab78ef001cad0b9cdfedb3aa3eaa6d4af/docs/StartGamePacket.html#L998-L1057).

Mojang’s type definition also identifies `PlayerPermissionLevel` as an `int8`, not an `int32`:

```json
{
  "x-minecraft-version": "1.26.50",
  "type": "string",
  "enum": ["Visitor", "Member", "Operator", "Custom"],
  "x-underlying-type": "int8"
}
```

See [`PlayerPermissionLevel.json`](https://github.com/Mojang/bedrock-protocol-docs/blob/7330880ab78ef001cad0b9cdfedb3aa3eaa6d4af/json/PlayerPermissionLevel.json#L1-L16).

### Conclusion

- The Mojang reference supports a **small signed variable-width integer type**, `varint8`.
- It does **not** support `zigzag32` for this field.
- For the valid permission values `0–3`, `varint8` and a raw `u8`/`int8` produce the same one-byte wire representation.
- Changing `PermissionLevel` to `zigzag32` would be incorrect for current protocol documentation and could produce a different wire format for values outside the normal `0–3` range.

This also explains the gophertunnel history: its September 2022 commit changed the field to `Varint32`, but current gophertunnel uses `Uint8`, which aligns more closely with Mojang’s current `varint8` reference. Therefore, I would recommend **not merging PR #1248 as written** unless it is narrowed to a specific historical protocol version and backed by version-specific Mojang schema evidence.

AnonymoDGH:
Closing this â€” @extremeheat's point is correct.

The field is a bounded byte encoded as a varint8 per Mojang's current protocol reference, and since only 32-bit varints are supported here, the existing type (closest to u8) is the right representation. My zigzag32 change was based on pmmp's `VarInt::readSignedInt`, but that doesn't reflect the actual wire format for this bounded field. Thanks for the correction.

PR #1262, opened 2026-08-29
Add quickMoveClickSendsEmptyItem, hasItemCommand and replaceItemSlotIsPrefixed features

u9g:
Three features needed by PrismarineJS/mineflayer#3979 so `lib/plugins/inventory.js` and the useChests test can use `bot.supportFeature` instead of hardcoded version comparisons:

- `quickMoveClickSendsEmptyItem` (1.12–1.16.5): for `actionIdUsed` versions the server accepts a click only if the packet's `item` matches the result of its own `slotClick`; for a quick move (mode 1) that is the moved stack up to 1.11 and empty from 1.12. From 1.17 the packet has no `item` field.
- `hasItemCommand` (1.17+): `/replaceitem` was replaced by `/item replace`.
- `replaceItemSlotIsPrefixed` (1.8–1.12.2): `/replaceitem` takes `slot.container.N` and a `minecraft:`-namespaced item name; 1.13 dropped the `slot.` prefix.

Boundaries verified against real servers on 1.8.8, 1.9.4, 1.11.2, 1.12.2, 1.16.5 and 26.1 locally and the rest of the 28-version matrix in that PR's CI.

### DIFF
--- a/data/pc/common/features.json
+++ b/data/pc/common/features.json
@@ -601,6 +601,14 @@
     ]
   },
   {
+    "name": "hasItemCommand",
+    "description": "/replaceitem was replaced by /item replace",
+    "versions": [
+      "1.17",
+      "latest"
+    ]
+  },
+  {
     "name": "indexesVillagerRecipes",
     "description": "gives a index for each trade in a villagers metadata",
     "versions": [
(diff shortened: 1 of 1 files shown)


rom1504:
Resolve conflicts

PR #1271, opened 2026-09-05
add bedrockR as a project that uses this

xFN10x:
Im also planning on reworking a bunch of internet stuff, so ill probably be using even more stuff at some point


extremeheat:
Cool, any specific reason you're using R?

extremeheat:
Ah nevermind it's in Java

xFN10x:
oh yea, the R in the name didn't really stand for anything at first, but i decided its short for bedrock(make)R

PR #1276, opened 2026-09-06
Align Bedrock 1.21.x block ranges with runtime state palettes

### LINKED ISSUE #1156: problem with the StateID of blocks, Bedrock
## Detailed description of a problem
Problem with a mismatch between the blockStates and blocksByStateId keys, the same index from the blockStates key points to another block in the blocksByStateId key. This issue concerns all other prismarine repositories whose users use the bedrock block register version 1.21.50 or higher.
## How to reproduce
```js
import PrismarineChunk from "prismarine-chunk"; // example on prismarine-chunk
import registry from "prismarine-registry"

const version = "1.21.50"

const startgamePacket = {/* there startgame packet from target server */}
const levelchunkPacket = {/* there levelchunkpacket from target server */}
const subchunkPacket = {/* there requested for levelchunk subchunk packet with y = 0 and diamond_block on x = 0, z = 0 from target server*/}

const reg = registry(`bedrock_${version}`)
reg.handleStartGame(startgamePacket)

const pchunkConstructor = PrismarineChunk(reg)
const PChunk = new pchunkConstructor({x: levelchunkPacket.x, z: levelchunkPacket.z})
await PChunk.networkDecodeNoCache(levelchunkPacket.payload, Number(levelchunkPacket.sub_chunk_count))
await PChunk.networkDecodeSubChunkNoCache(Number(subchunkPacket.dy), subchunkPacket.payload)

PChunk.getBlock({x: 0, y: 0, z: 0}) // Outpus: name: 'end_bricks' Instead name: 'diamond_block'
```

## Additional context
I found this problem inside [prismarine-registry](https://github.com/PrismarineJS/prismarine-registry/issues/52), this problem extends to bedrock versions 1.21.5x and higher.

### Example of method loadRuntimeIds which uses problem keys
```js
function loadRuntimeIds () {
    data.blocksByRuntimeId = {}
    for (let i = 0; i < data.blockStates.length; i++) {
      // i = 11996, data.blockStates[i].name = 'air'
      data.blocksByRuntimeId[i] = { stateId: i, ...data.blocksByStateId[i] } // data.blocksByStateId[i].name = "crimson_hanging_sign"
    }
  }
```
---

Pix3lPirat3:
- add matching `blocks.json` tables for Bedrock 1.21.20, 1.21.30, and 1.21.50;
- point those releases at their matching block tables;
- validate complete, non-overlapping state ranges, in-range defaults, and matching block names.

The affected versions had distinct `blockStates.json` palettes but reused incompatible `blocks.json` tables. This caused state IDs such as `defaultState` to resolve to the wrong block. The tables use the same pinned PMMP BedrockData palettes validated by the retained official-BDS probes.

Fixes #1156.

## Tests

- mapping invariant: 30 unique block/state pairs passing
- full minecraft-data suite: 1,896 passing, 1 expected pending
- BDS 1.21.20, 1.21.30, and 1.21.50 world/block probes passing
- dependency-integrated BDS world/block matrix: 51 passing, 0 failing, 2 unavailable


extremeheat (approved):

tools/js/test/bedrock_block_states.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-env mocha */
```
extremeheat, line 1: Wonder if we could enforce this at a schema level--define some items as unique like in SQL that we can catch generically via some tests. Would help catch this kind of error more broadly for rest of the data

extremeheat:
Merging to make way for #1297

PR #1277, opened 2026-09-06
bedrock: fix legacy conditional packet payloads

Pix3lPirat3:
- select the 1.16.201 and 1.16.210 boss-event payload using the mapped `show_bar` value;
- decode score identity fields for `change` entries instead of `remove`;
- regenerate protocol JSON and add regression coverage for both versions.

The released schemas decode boss IDs but discard title/progress, while score changes lose the identity fields required to associate scores with players or custom names. Bedrock 1.16.220 already uses the corrected boss selector.

### DIFF
--- a/data/bedrock/1.16.201/proto.yml
+++ b/data/bedrock/1.16.201/proto.yml
@@ -1283,7 +1283,7 @@ packet_boss_event:
    _: type?
       if register_player or unregister_player:
          player_id: zigzag64
-      if show:
+      if show_bar:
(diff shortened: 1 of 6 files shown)


extremeheat (approved):
Good catch, seems like we should have a sanity check of sorts to prevent switching on undefined map value

extremeheat:
Thanks, this was done via #1289 which also fix mcpc data, but we should move the validation to protodef-validator

PR #1278, opened 2026-09-07
pc/26.1: map every InteractionHand field to main_hand/off_hand

u9g:
In pc/26.1, `use_entity.hand` is a `mapper` (`main_hand`/`off_hand`), while the other `InteractionHand` fields (`open_book`, `arm_animation`, `block_place`, `use_item`) are plain varints. That means a client has to write `'main_hand'` for one packet and `0` for the rest.

This PR keeps the named form and applies it to all five fields. That fits the move toward enum mappers in #1295, and it avoids the silent regression flagged in review: with a plain varint, ProtoDef writes `off_hand` as `00`.

- `data/pc/latest/proto.yml`: `open_book`, `arm_animation`, `block_place` and `use_item` get the same `main_hand`/`off_hand` mapping as `use_entity`; `protocol.json` is regenerated.
- `edit_book.hand` is left as a varint because vanilla's `ServerboundEditBookPacket` carries a hotbar slot index there, not an `InteractionHand`.
- `tools/js/test/protocolHands.js` (new `protodef` devDependency): for each of the five fields, serializes both `main_hand` → `00` and `off_hand` → `01`, and reads `01` back as `off_hand`.

Callers that write numeric ids to the four newly mapped fields will now get a serializer error, not a wrong hand.

### DIFF
--- a/data/pc/26.1/protocol.json
+++ b/data/pc/26.1/protocol.json
@@ -10852,16 +10852,7 @@
             },
             {
               "name": "hand",
-              "type": [
-                "mapper",
-                {
-                  "type": "varint",
-                  "mappings": {
-                    "0": "main_hand",
-                    "1": "off_hand"
-                  }
-                }
-              ]
+              "type": "varint"
             },
             {
               "name": "location",


u9g:
Confirmed on a live 26.1 server (JartexNetwork, 1.8 backend behind ViaVersion). Worth noting how it fails, because it does not look like a serializer error from the outside: the throw happens in `Serializer._transform`, which errors the write stream and closes the socket without a `kicked` or `error` event reaching the bot. `bot.entity` still reads and `physicsEnabled` is still true, so a control loop keeps driving a bot that has sent nothing for minutes.

With this change `bot.activateEntity` on a shop villager opens the window and the purchase goes through.

u9g:
Context, in case you'd rather go the other way: mineflayer has since been fixed to write the mapper name (`hand: 'main_hand'`) in PrismarineJS/mineflayer#4066 and #4076, so it no longer depends on this change.

What's left is the inconsistency itself — inside pc/26.1, `use_entity.hand` is a mapper while `arm_animation.hand`, `block_place.hand`, `use_item.hand` and `edit_book.hand` are plain varints, and all five are the same `InteractionHand` enum in vanilla. Whichever direction you prefer is fine by me; making the other four mappers instead would resolve it just as well, and I'm happy to redo the PR that way.

data/pc/26.1/protocol.json
```diff
@@ -10852,16 +10852,7 @@
...
-                "mapper",
-                {
-                  "type": "varint",
-                  "mappings": {
-                    "0": "main_hand",
-                    "1": "off_hand"
-                  }
-                }
-              ]
+              "type": "varint"
```
rom1504, line 10855: **Astra agent review — AI-generated, not manually written by the maintainer.**

Removing this mapper silently changes an existing `hand: 'off_hand'` write from byte `01` to `00` (main hand). I reproduced this with ProtoDef 1.19.0 using the changed field type: its varint writer coerces both names to zero, so this is a wrong-hand interaction rather than a useful migration error. Could we keep the named representation and adapt numeric callers, or supply an explicit compatibility path before removing it? Please cover both named hands in a serializer regression test; testing only main hand hides the regression.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current diff and discussion and locally reproduced named-value serialization before/after with ProtoDef 1.19.0. The inline finding concerns silent behavior changes for callers using the existing named API; no live-server test was run.

data/pc/26.1/protocol.json
u9g: Agreed, reworked in the other direction (237e8a8): `use_entity.hand` keeps its mapper, and `open_book`, `arm_animation`, `block_place` and `use_item` get the same `main_hand`/`off_hand` mapping, so the named form is used for every `InteractionHand` field in 26.1. `tools/js/test/protocolHands.js` serializes both named hands for each field and checks `off_hand` → `01`. A plain varint writes `00` there, so the test catches the regression you found.

u9g:
Correction to my earlier comment: PrismarineJS/mineflayer#4066 and #4076 were closed without merging, so mineflayer master still writes `hand: 0` to `use_entity` on 26.1. Writing names there (and now also for `arm_animation`, `block_place` and `use_item`) is pending in mineflayer.

rom1504 (commented):

Looks ready to merge from this review. The revised direction preserves use_entity's mapper and adds matching names to the other four hand fields in both YAML and generated JSON. The five added tests pass. Separately, using the complete candidate schema in an isolated process with NMP 1.68.0 / ProtoDef 1.19.0, I encoded and decoded all five full packets: named and numeric main/off hands produce identical bytes (10 pairs), and off_hand remains 1. This avoids the earlier interpreter-versus-production confusion. Current CI is green; no live-server action was claimed.

Skills used: `prismarine-protocol-data-review` traced the maintained data through its production codec/consumer; `prismarine-review` checked current revisions, CI and existing discussion to avoid duplicate findings.

PR #1287, opened 2026-09-10
[Java] Fill data gaps from vanilla sources: 26.2, 1.13.1 palettes, 1.14.2, windows, commands, 1.9.1

milutinke:
Hey,

I kept tripping over holes in this dataset while building a .NET client (UMPK) against it, so instead of working around them I figured I'd send the fixes back. Everything below comes from vanilla itself: jars, data generator reports, decompiled sources, and live packet captures where it mattered.

What was wrong:

The 1.13.1 entry reuses all of 1.13's blocks, items, and other. Real 1.13.1 has 598 blocks and 789 items. Five dead coral blocks and items showed up, live corals and the conduit got waterlogged, TNT got unstable, and that shifts every state id from TNT onward and every item id from 443 onward. Anyone decoding 1.13.1 with this data was reading garbage past those points. I replaced those files with the real palettes from the 1.13.1 data generator reports. The pre-releases had to move too since the dead corals postdate them.

What was missing:

Protocol 776 (26.2) had nothing at all, not even a dataPaths entry. I added 19 files. Language is byte-exact against Mojang's en_us (8123 entries), blocks and items check out against the report palettes (1196 blocks, 1537 items), and protocol.json follows the usual generated process with each of the 8 wire changes diffed against the 26.2 sources.

Protocol 485 (1.14.2) was also completely absent. Full entry now, 676 blocks and 876 items, all matching. One detail: the trade list packet on 485 has no canRestock field (that arrives at 490), so I left it out.

Protocol 108 (1.9.1) was only a pre2 stub. I completed the entry, confirmed the 1.9 final data stands for it, and the language table is byte-exact (2855 entries). Side note: the "1.9" language file looks like it's actually the 1.9.4 table minus two keys. Left it alone as out of scope.

windows.json: everything from 1.16.2 up still points at the 1.16.1 file, so no crafter or anything newer. I added a 1.20.3 windows file with the crafter (checked against the decompiled CrafterMenu in four vanilla versions; it's the only menu added since 1.16.1) and repointed 15 versions. Caveat: the crafter property names are my own labels for verified container indices.

commands.json: everything newer than 1.20.3 pointed at the 1.20.3 tree. I converted the actual vanilla command reports for 1.20.5, 1.21.5, 1.21.6, 26.1, and 26.2, node for node (1763 to 2091 nodes each). Versions in between borrow the nearest older exact tree.

Two drive-by fixes in the new files: trial_spawner_detected_player is actually trial_spawner_detection, and zomie should be zombie.

Sources, all vanilla: Mojang server jars and their data generator reports (blocks, items, registries, commands), the vanilla en_us language files, decompiled sources (Blocks, CrafterMenu, packet classes), and live packet captures for wire layouts.

Validation: full mocha suite passes, 1917 tests, zero failures, including schema checks on everything new. Every file was cross-checked against the UMPK dataset, which is extracted straight from vanilla.

Knowingly left alone: 1.13.1 recipes still borrow from 1.13, effects/enchantments/instruments on 26.2 are borrowed (those registries didn't change, verified), and two stale packet names (0x40 spectate, block_place) that predate this PR.

Happy to split this up if smaller pieces are easier to review.


extremeheat:
Cool, but there are alot of separate changes in here mixed in that make it hard to review. Can you please break into separate PRs individually fixing the relevant issues?

If you used LLM you can ask it to break up the diff into multiple PRs

DallasCarraher:
To help get 26.2 support landed sooner, I split the 26.2-specific slice of this PR out into #1298 — just the `data/pc/26.2/*` files, the new `1.20.3/windows.json`, and the `latest/proto.yml` bump, without the unrelated 1.13.1/1.14.2/1.9.1/commands-refresh changes bundled in here. Full test suite passes there. Figured a smaller, single-purpose PR might be easier to review independently of the rest of this changeset — happy to fold it back in if you'd rather keep it all together.

PR #1291, opened 2026-09-11
fix recipe serializer IDs in 1.20.5 and 1.21.1

VasilisDragon:
follow-up to #1286. `crafting_special_banneraddpattern` isn't registered in either version; adding it shifted the existing IDs, so a smelting test packet decoded as repairitem.

restores the registry order and removes the stale switch case.

checked against the official client registries and 50 synthetic recipe fixtures with the compiled codec. the local full data suite hit the existing 40s timing assertion on both master and this branch.


tools/js/test/fixtures/recipe-serializers.json
extremeheat, line 1: Remove

tools/js/test/recipeSerializerIds.js
extremeheat, line 1: Remove

tools/js/test/fixtures/recipe-serializers.json
VasilisDragon: Removed both, thanks.

PR #1219, opened 2026-08-11
🎈 Add Minecraft pc 26.2 data

rom1504bot:
This automated PR sets up the relevant boilerplate for Minecraft pc version 26.2. Fixes https://github.com/PrismarineJS/minecraft-data/issues/1218.

Related:
- Issue: https://github.com/PrismarineJS/minecraft-data/issues/1197
- Protocol Version: 776

* You can help contribute to this PR by opening a PR against this <code branch>pc_26_2</code> branch instead of <code>master</code>.

Issue #1197, opened 2026-06-16
Support Minecraft PC 26.2


rom1504bot:
A new Minecraft Java Edition version is available (as of 2026-06-16T12:03:33+00:00), version **26.2**
## Official Changelog
* https://feedback.minecraft.net/hc/en-us/sections/360001186971-Release-Changelogs
## Protocol Details
(I will close this issue automatically if "26.2" is added to data/pc/common/versions.json on "master")
<table>
  <tr><td><b>Name</b></td><td>26.2</td>
  <tr><td><b>Protocol ID</b></td><td>776</td>
  <tr><td><b>Release Date</b></td><td>2026-06-16T12:03:33+00:00</td>
  <tr><td><b>Release Type</b></td><td>release</td>
  <tr><td><b>Data Version</b></td><td>4903</td>
  <tr><td><b>Java Version</b></td><td>25</td>
</table>

<hr/>
🤖 I am a bot, I check for updates every 2 hours without a trigger. You can close this issue to prevent any further updates.

p-romeo:
> **Disclosure: this work was done largely with AI coding tools (Claude Code).** Verified against a live 26.2 server and the 26.2 server jar, and covered by an offline test suite, but please review it as AI-assisted work.

I've been running a bot on 26.2 (protocol 776) against a vendored copy of this data, and hit two `protocol.json` bugs worth fixing before this branch lands. Both are verified against the jar.

**1. Serverbound play packet IDs are off by one from `0x26` onward.**

The generator misses `PingPacketTypes.SERVERBOUND_PING_REQUEST` at `0x26` — it lives outside `GamePacketTypes`, so walking that class alone doesn't find it. Every subsequent ID shifts down by one.

Symptom is a kick on the first interaction, because `block_dig` gets encoded as `player_abilities` and `use_item` as `use_item_on`.

Corrected against `GameProtocols.SERVERBOUND_TEMPLATE` in the 26.2 jar; IDs then match exactly through `custom_click_action` at `0x44`, including `teleport_to_entity` at `0x40`.

**2. Advancement icons and recipe item-stack displays need an `ItemStackTemplate` type**, and the data-component ID list is missing the 26.2 additions.

[...]

PR #1267, opened 2026-08-30
pc: fix intangible_projectile NBT and 26.1 item stack templates

u9g:
Two protocol fixes found by PrismarineJS/prismarine-item#184's live vanilla tests.

### intangible_projectile is NBT, not void (1.21.5–26.1)

The component has no dedicated network codec, so vanilla syncs it through its NBT codec — an empty compound (`0a 00`) on the wire. 1.20.5–1.21.4 correctly had `anonymousNbt` and 1.21.5 onward regressed to `void`, leaving two bytes unread in any slot carrying the component (protodef PartialReadError on `set_slot`). `unbreakable` and `glider` stay `void`: they have real unit stream codecs and decode cleanly today.

### 26.1 sends non-empty stacks as item stack templates (item id first)

26.1 changed the non-optional stack codec to item id, count, components; the optional count-first `Slot` form (`set_slot`, `window_items`, cursor, equipment, merchant offers, entity metadata) is unchanged. Before this fix, nested stacks decode with `itemId`/`itemCount` swapped — giving `bundle[bundle_contents=[{id:stone,count:3},{id:dirt}]]` parsed as item 3 ×1 and item 1 ×28.

This adds an `ItemStackTemplate` type and uses it where vanilla uses the template codec (cross-checked against MCProtocolLib's `readItemStackTemplate` call sites):

- `use_remainder`, `charged_projectiles`, `bundle_contents` components
- `container` component — now a **prefixed-optional** template per entry, preserving empty slots
- the `item` particle
- the `item_stack` slot display
- the advancement display icon

### Verification

Each site was checked against real vanilla servers (1.21.5, 1.21.6, 1.21.8, 1.21.9, 1.21.11, 26.1) by giving items carrying each component, granting an advancement and spawning item particles, then confirming the decoded ids and counts match what was given, with zero parse errors. On 26.1: stone(1)×3 + dirt(28)×1 bundle, shulker `container` decoding `[null, diamond(899)×5]`, advancement icons and `minecraft:item` particle all correct. prismarine-item#184's 78 live window_click hash tests still pass against 1.21.5/1.21.11 with these files, now with no PartialReadErrors.

`protocol.json` files are regenerated from the edited `proto.yml`s via `npm run build`; `tools/js` tests pass (the only local failure is the 40s suite-duration gate, from running Minecraft servers concurrently on the same machine).

PR #1298, opened 2026-09-14
Add Minecraft PC 26.2 data (protocol 776)

DallasCarraher:
## Summary
Registers Minecraft PC 26.2 (protocol 776), closing the gap tracked in #1197 and the stalled scaffold in #1219.

This is the 26.2-specific slice of #1287, pulled out so it can be reviewed and merged on its own:
- All 19 `data/pc/26.2/*.json` files (blocks, items, entities, recipes, commands, loot tables, language, protocol.json regenerated from proto.yml, etc.), taken verbatim from #1287, sourced there from Mojang's 26.2 data generator reports/decompiled sources and cross-checked against a third-party extraction.
- `data/pc/1.20.3/windows.json` (new file, also from #1287) adding the `crafter_3x3` menu that's been missing from `windows.json` since 1.20.3, so 26.2's `windows` dataPath entry is accurate instead of falling back to the stale 1.16.1 file.
- `data/pc/latest/proto.yml` bumped from 26.1 to 26.2 with the 8 wire-format diffs (sessionId, onlineMode, team packet reorder, spectator_action rename, sulfur_cube slot component, particle registry additions, etc.).
- Registration: `data/dataPaths.json`, `data/pc/common/versions.json`, `README.md`.

Deliberately **not** included from #1287: the unrelated 1.13.1/1.14.2/1.9.1 fixes and the broader commands.json refresh for older already-released versions — those are separate concerns better reviewed on their own.

`effects`/`enchantments`/`instruments` are borrowed from 26.1 since those registries are unchanged in 26.2 (per #1287's own notes).

## Test plan
- [x] `npm test` in `tools/js` (mocha + ajv schema validation): 1887 passing, 1 pending, 0 failing
- [x] `npm run lint` (standard): clean
- [x] Live-tested against a real 26.2 server with mineflayer/node-minecraft-protocol


DallasCarraher:
Live-tested this against a real vanilla 26.2 server (official Mojang server.jar, protocol 776, offline mode, verified sha1).

**Protocol-level test (`minecraft-protocol`, using this branch's data via a `node-minecraft-data` submodule swap):**
- TCP connect → login → 82 packets parsed, including a full chunk packet → 0 parse errors.
- Confirms `protocol.json` correctly decodes the real 26.2 wire format end to end.

**mineflayer:**
`mineflayer@latest` and `prismarine-chunk@latest` both hardcode their own per-major-version allowlists independent of minecraft-data (`lib/version.js`'s `testedVersions`, and `chunkImplementations.pc` in prismarine-chunk). Both needed a one-line `26.2` entry (aliased to the same 1.18-family chunk impl 26.1 uses, since nothing in this PR changes the chunk section format) before mineflayer would even attempt a connection — expected, unrelated follow-up PRs in those repos once this lands.

With those two bumped locally, I also hit `prismarine-physics@1.11.1` throwing `No liquid gravity settings, have you made sure the liquid gravity features are up to date?` — but this isn't 26.2-specific: `independentLiquidGravity`/`proportionalLiquidGravity` aren't defined in `features.json` for *any* version on current master, so this reproduces identically on 26.1 or 1.21.11 with the currently published prismarine-physics. Pre-existing cross-repo gap, not something this PR needs to fix.

Summary: the actual minecraft-data changes in this PR check out against a live server. The mineflayer-side blockers are separate, expected follow-ups in sibling repos.

DallasCarraher:
Update: getting this working end-to-end against a real mineflayer bot (pathfinder, collectblock, pvp, tool, armor-manager) surfaced four downstream gaps in sibling repos — none blocking this PR itself, but each needed for a full 26.2 bot to actually work once this data lands. Opened fixes for all of them, each independently tested against this branch:

- [PrismarineJS/prismarine-chunk#333](https://github.com/PrismarineJS/prismarine-chunk/pull/333) — chunk implementation mapping was missing for 26.2
- [PrismarineJS/prismarine-physics#143](https://github.com/PrismarineJS/prismarine-physics/pull/143) — `proportionalLiquidGravity` feature list was missing 26.2 (pre-existing gap, not 26.2-specific in cause — reproduces on 26.1 too with the currently published package)
- [PrismarineJS/node-minecraft-protocol#1529](https://github.com/PrismarineJS/node-minecraft-protocol/pull/1529) — server-side login success packet needed the new `sessionId` field
- [PrismarineJS/mineflayer#4121](https://github.com/PrismarineJS/mineflayer/pull/4121) — 26.2 added to `testedVersions` (only mergeable once this PR ships in a release, since it gates support at require-time)

With all four applied locally on top of this branch: mineflayer's internal test suite passes 39/39 (1 pending) for 26.2, node-minecraft-protocol's full suite passes 6787/6789 (2 pre-existing unrelated failures, zero attributable to 26.2), and a live bot with the full plugin stack spawned and pathfound successfully against a real vanilla 26.2 server.

extremeheat:
https://github.com/PrismarineJS/minecraft-data/pull/1219 isn't stalled, it's the automatically generated boilerplate for supporting new version. One branch allows multiple people to work toward it vs needing one large PR. We can merge PRs toward that branch as there already PRs testing against it in node-minecraft-protocol and mineflayer CI wise


So we can likely point this PR against `pc_26_2` branch

extremeheat:
I updated pc_26_2 branch to latest master

Can you rebase this on that branch/cherry pick commits on new checkout of that branch to resolve conflict? Thanks

DallasCarraher:
Done — rebased onto `pc_26_2` and retargeted this PR's base branch there. Force-pushed the same branch rather than opening a new PR, so the history/comments stay put.

One thing worth flagging from the rebase: `pc_26_2`'s scaffold `protocol.json` had the same `sessionId`/`onlineMode` fields already patched in (nice, matches what I'd found independently), but it's otherwise still the stale copy-of-26.1 scaffold — missing the other wire changes this PR adds (team packet reorder, `spectator_action` rename, `sulfur_cube` slot component, particle registry additions, etc.), all diffed against the real 26.2 sources and live-tested against a vanilla server. Since the two files overlapped on the same lines, git flagged it as a conflict rather than merging cleanly — resolved by taking this PR's complete `protocol.json` wholesale. Also had to deduplicate a leftover second `"26.2"` key in `dataPaths.json` that the auto-merge produced (one from the scaffold's minimal entry, one from this PR's full one) — kept this PR's complete entry.

Full test suite still green post-rebase: 1887 passing, 1 pending, 0 failing.

data/pc/latest/proto.yml
```diff
@@ -3752,9 +3760,9 @@
       text2: string
       text3: string
       text4: string
-   # MC: ServerboundSpectateEntityPacket
-   packet_spectate_entity:
-      entityId: varint
+   # MC: ServerboundSpectatorActionPacket
```
extremeheat, line 3770: Look into if this is same packet renamed or new packet

data/pc/latest/proto.yml
```diff
@@ -456,6 +457,14 @@
                nbtData: anonymousNbt
                ticksInHive: varint
                minTicksInHive: varint
+         if sulfur_cube_content:
```
extremeheat: Looks like this is supposed to refer to some item data type . But being redefined in-line

data/pc/latest/proto.yml
DallasCarraher: Verified — it's a genuine rename, not a new packet alongside the old one. Decompiled the real 26.2 server jar (`versions/26.2/server-26.2.jar`, unpacked from the official sha1-verified `server.jar`):

- `net/minecraft/network/protocol/game/ServerboundSpectatorActionPacket.class` exists: a `Record` with a single field `private final java.util.OptionalInt spectateEntityId`, registered as `GamePacketTypes.SERVERBOUND_SPECTATOR_ACTION` under the string id `spectator_action`, dispatched via `ServerGamePacketListener.handleSpectatorAction`.
- No `ServerboundSpectateEntityPacket` class exists anywhere in the jar (searched the whole jar; the only remaining "Spectate"-named class is the unrelated `SpectateCommand`).

So this PR's existing encoding (`entityId?: varint`, optional) already matches the real packet exactly — no change needed here. Pushed a commit (3833a145) documenting this in the message.

data/pc/latest/proto.yml
DallasCarraher: Good catch — this is exactly `ItemStackTemplate` (added separately in #1267, already on this branch's base). Fixed to `if sulfur_cube_content: ItemStackTemplate`.

Verified against the real 26.2 server jar: `net.minecraft.world.item.component.SulfurCubeContent` is a `Record` with a single field, `net.minecraft.world.item.ItemStackTemplate absorbedBlockItemStack` — no array wrapper, and Mojang's own internal type is literally named `ItemStackTemplate` too.

While fixing this I also found that rebasing this branch onto `pc_26_2` had silently dropped #1267's `ItemStackTemplate` fix throughout `data/pc/latest/proto.yml` — git merged both sides' overlapping edits to this file without flagging a conflict, but the result dropped #1267's hunks. Restored all of it and regenerated `data/pc/26.2/protocol.json`. Pushed in 3833a145; full test suite still green (1887 passing, 0 failing).

data/pc/latest/proto.yml
```diff
@@ -2962,12 +2977,8 @@
       x: f64
       y: f64
       z: f64
-      dx: f64
-      dy: f64
-      dz: f64
-      yaw: f32
-      pitch: f32
-      flags: PositionUpdateRelatives
+      yaw: i8
```
VasilisDragon: **AI agent review — manually approved by Vasilis.**

The [26.2 teleport codec](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/network/protocol/game/ClientboundTeleportEntityPacket.java#L32) still includes velocity and relative flags, with float yaw/pitch in [PositionMoveRotation](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/PositionMoveRotation.java#L56). Removing those fields and using byte angles changes the wire layout.

A 62-byte payload produced by the vanilla codec is fully consumed by the base, but this definition consumes 28 bytes and decodes the angles incorrectly. All three vanilla teleport fixtures fail with this change. Please retain the existing layout and regenerate the JSON.

data/pc/latest/proto.yml
```diff
@@ -3752,9 +3763,9 @@
       text2: string
       text3: string
       text4: string
-   # MC: ServerboundSpectateEntityPacket
-   packet_spectate_entity:
-      entityId: varint
+   # MC: ServerboundSpectatorActionPacket
+   packet_spectator_action:
+      entityId?: varint
```
VasilisDragon: **AI agent review — manually approved by Vasilis.**

The rename matches vanilla, but the [packet codec](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/network/protocol/game/ServerboundSpectatorActionPacket.java#L26) uses [OPTIONAL_VAR_INT](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/network/codec/ByteBufCodecs.java#L111): `0` for absent, otherwise `entityId + 1`, without a separate presence byte.

For entity 42, vanilla writes `2b`; `entityId?: varint` writes `01 2a` and cannot read the vanilla payload. The absent fixture passes, but all six present-ID fixtures fail. Please represent the sentinel-based encoding here and regenerate the JSON.

VasilisDragon (commented):
**AI agent review — manually approved by Vasilis.**

Reviewed `3833a145` against the PR base. Checked the changed protocol definitions and data against hash-verified vanilla 26.2 artifacts, executed selected vanilla codecs, and compared baseline/candidate decoding, encoding and loader output. The inline findings cover two wire-format issues and three entity/attribute data corrections. The earlier ItemStackTemplate concern is addressed at this revision.

Generation produced no drift, including a repeated candidate build. Existing tooling and packet tests retained failures reproduced on the base. Interpreted codec diagnostics needed a separate adapter for an existing NBT-registration typo. No live vanilla server testing was performed, and the selected cycle tests had no recorded packets.

data/pc/26.2/entities.json
```diff
@@ -0,0 +1,4440 @@
...
+      "ticks_frozen",
+      "living_entity_flags",
+      "health",
+      "effect_particles",
+      "effect_ambience",
+      "arrow_count",
+      "stinger_count",
+      "sleeping_pos",
+      "mob_flags",
+      "size",
```
VasilisDragon, line 3639: **AI agent review — manually approved by Vasilis.**

These metadata keys need the new inherited fields. In 26.2, [AbstractCubeMob extends AgeableMob](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/monster/cubemob/AbstractCubeMob.java#L42), whose [accessors](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/AgeableMob.java#L274) occupy IDs 16 (`baby`) and 17 (`age_locked`). `size` moves to 18. [SulfurCube then allocates](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/monster/cubemob/SulfurCube.java#L859) `max_fuse` at 19 and `from_bucket` at 20.

Currently, a size update at ID 18 maps to `max_fuse` for sulfur_cube and has no metadata key for slime or magma_cube. Please update all three entries, preserving the accessor allocation order.

data/pc/26.2/entities.json
```diff
@@ -0,0 +1,4440 @@
...
+      "mob_flags",
+      "flags"
+    ]
+  },
+  {
+    "id": 11,
+    "internalId": 11,
+    "name": "bee",
+    "displayName": "Bee",
+    "width": 0.7,
```
VasilisDragon: **AI agent review — manually approved by Vasilis.**

Bee dimensions changed to width `0.55`, height `0.5` in 26.2; [EntityTypes registers `sized(0.55F, 0.5F)`](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/EntityTypes.java#L345). These are still the 26.1 dimensions, so loader consumers receive the old base hitbox. Please update both values.

data/pc/26.2/attributes.json
```diff
@@ -0,0 +1,282 @@
...
+    "name": "jumpStrength",
+    "resource": "minecraft:jump_strength",
+    "min": 0.0,
+    "max": 32.0,
+    "default": 0.41999998688697815
+  },
+  {
+    "name": "knockbackResistance",
+    "resource": "minecraft:knockback_resistance",
+    "min": 0.0,
```
VasilisDragon: **AI agent review — manually approved by Vasilis.**

The minimum should be `-2.0` in 26.2. [Vanilla registers this attribute](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/world/entity/ai/attributes/Attributes.java#L74) with default `0.0`, minimum `-2.0` and maximum `1.0`. Keeping `0.0` excludes valid negative values for callers using these bounds. Please update the minimum.

data/pc/latest/proto.yml
DallasCarraher: Confirmed and fixed — this is the exact same failure mode as the ItemStackTemplate issue: rebasing onto `pc_26_2` also silently dropped #1273's 1.21.2+ teleport layout fix (velocity, f32 rotation, relative flags) from `data/pc/latest/proto.yml`, reverting it to the pre-1.21.2 byte-angle layout. Restored it, and confirmed it now matches `packet_position` (the player-position packet) field-for-field, per the original fix's own note that `ClientboundTeleportEntityPacket` uses the same `PositionMoveRotation` layout. Regenerated `protocol.json`, full test suite green (1887 passing, 0 failing). Pushed in 2003e226.

data/pc/latest/proto.yml
DallasCarraher: Confirmed and fixed. Decompiled `ByteBufCodecs.class`: `OPTIONAL_VAR_INT` is built from a raw `VarInt.read`/`VarInt.write` codec (`ByteBufCodecs$6`, no separate presence byte at all) wrapped in `.map(...)` to convert to/from `OptionalInt` — exactly the sentinel encoding you described (0 = absent, n = id + 1).

Changed `entityId?: varint` to `entityId: optvarint` — this codebase already has that exact alias (`optvarint: varint`) for this exact vanilla pattern, used by entity metadata's `optional_block_state`/`optional_unsigned_int`. Regenerated `protocol.json`, full test suite green. Pushed in 2003e226.

data/pc/26.2/entities.json
DallasCarraher: Confirmed and fixed. Decompiled the accessor definition order directly: `AgeableMob.<clinit>` defines `DATA_BABY_ID` then `AGE_LOCKED` (both booleans, indices 16/17 after the 16 inherited from Entity/LivingEntity/Mob), `AbstractCubeMob.<clinit>` defines one `ID_SIZE` (int, index 18), and `SulfurCube.<clinit>` defines `MAX_FUSE` (int) then `FROM_BUCKET` (boolean) — indices 19/20, in that order.

slime and magma_cube were missing `baby`/`age_locked` entirely, and sulfur_cube had `from_bucket`/`max_fuse` both mis-positioned and swapped. Fixed all three to `..., mob_flags, baby, age_locked, size, max_fuse, from_bucket`. Pushed in 2003e226.

data/pc/26.2/entities.json
DallasCarraher: Confirmed and fixed. Decompiled `EntityTypes.<clinit>`: `BEE = register(..., Builder.of(Bee::new, MobCategory.CREATURE).sized(0.55f, 0.5f)...)`. Was still the 26.1 values (0.7/0.6). Fixed. Pushed in 2003e226.

data/pc/26.2/attributes.json
DallasCarraher: Confirmed and fixed. Decompiled `Attributes.class`: `KNOCKBACK_RESISTANCE = register("knockback_resistance", new RangedAttribute(..., 0.0, -2.0, 1.0))` (default, min, max). Was `min: 0.0`. Fixed to `-2.0`. (Double-checked I wasn't confusing this with the separate `explosion_knockback_resistance` attribute, which genuinely does have min 0.0 — different attribute entirely.) Pushed in 2003e226.

DallasCarraher:
Addressed all five findings from the latest review (2003e226) — verified each independently by decompiling the real 26.2 server jar with CFR before fixing, rather than taking the review at face value:

1. `packet_entity_teleport` — the `pc_26_2` rebase had also silently dropped #1273's 1.21.2+ layout fix (same failure mode as the ItemStackTemplate regression from the last round). Restored; now matches `packet_position`'s layout field-for-field, as expected.
2. `packet_spectator_action.entityId` — confirmed vanilla's `OPTIONAL_VAR_INT` is a raw sentinel varint with no presence byte, not protodef's standard `option`. Changed to `optvarint` (the existing alias already used for this exact pattern elsewhere).
3. `entities.json` slime/magma_cube/sulfur_cube — confirmed via `SynchedEntityData.defineId` call order in `AgeableMob`/`AbstractCubeMob`/`SulfurCube` bytecode. Added missing `baby`/`age_locked`, fixed sulfur_cube's swapped `max_fuse`/`from_bucket`.
4. bee dimensions — confirmed `EntityTypes.BEE.sized(0.55f, 0.5f)`; was still 26.1's 0.7/0.6.
5. `knockbackResistance` attribute min — confirmed `RangedAttribute(0.0, -2.0, 1.0)`; was 0.0.

Full test suite green throughout (1887 passing, 1 pending, 0 failing), and re-verified live connectivity against a real 26.2 server after each round of fixes.

data/pc/latest/proto.yml
extremeheat: No, I meant a rename would be semantically same thing under different name (we should not rename packets like this for compatibility reasons). New packet would be semantically different

DallasCarraher: Confirmed — no rename actually happened here. `data/pc/26.1/proto.yml` still has its own `packet_spectate_entity` untouched; `latest`/26.2 adds `packet_spectator_action` as a brand-new identifier (distinct field shape too: optional/sentinel-encoded vs. mandatory). My earlier commit message calling it "a rename" was just imprecise wording — the actual data change is already a new packet, not an in-place rename. Nothing further to change here.

data/pc/latest/proto.yml
```diff
@@ -1019,15 +1028,16 @@
          - shriek
          - egg_crack
          - dust_plume
-         - trial_spawner_detected_player
-         - trial_spawner_detected_player_ominous
+         - trial_spawner_detection
+         - trial_spawner_detection_ominous
```
extremeheat: Backport to prev relevant versions or preserve old naming

Pix3lPirat3:
Heads-up: `data/pc/26.2/blocks.json` here has `material: incorrect_for_wooden_tool` on 108 blocks (ores, obsidian, metal blocks), which makes `prismarine-block` `digTime()` ~8x too slow for them - the bug #1307 fixes for 1.20.5-26.1.

It comes from PrismarineJS/minecraft-data-generator#77's 26.2 module lacking the fix from PrismarineJS/minecraft-data-generator#86. Regenerating once that lands, or applying the same `incorrect_for_wooden_tool -> mineable/pickaxe` mapping to those 108 blocks, would keep 26.2 from regressing.

extremeheat:
Merging into 26.2 branch to remove PR congestion

VasilisDragon:
One remaining follow-up for `pc_26_2`: could we regenerate the embedded login registries from 26.2? [`loginPacket.json`](https://github.com/PrismarineJS/minecraft-data/blob/9e2f9adf4a95f7c6f747b915044080ca94cc7a3b/data/pc/26.2/loginPacket.json#L11956) is unchanged from `2003e226`, which I tested in the earlier review.

Vanilla 26.2's `Enchantment.DIRECT_CODEC` rejects eight entries: `bane_of_arthropods`, `channeling`, `impaling`, `looting`, `lunge`, `power`, `punch` and `smite`. They still use older entity-predicate structures. For `bane_of_arthropods`, changing just the two predicate keys from `type` to `minecraft:entity_type` makes it decode and match the official resource. The other entries need their current structures too. Source: [`EntityPredicate`](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/advancements/predicates/entity/EntityPredicate.java#L130) and [registered subpredicate names](https://github.com/extremeheat/extracted_minecraft_data/blob/912ae031240da395051acae238a3f60d733353de/client/net/minecraft/advancements/predicates/entity/EntitySubPredicates.java#L12).

All 43 official enchantment resources passed the same harness and registry context. This is inherited stale data, not a regression from the latest fixes. NMP [uses these registries by default for servers](https://github.com/PrismarineJS/node-minecraft-protocol/blob/0ebdd5120eca8840e3ffd6597cd4e83732b59c9a/src/createServer.js#L49); these were native-codec checks, not a live vanilla-client connection.

PR #1302, opened 2026-09-16
pc: map 1.21.5 team nameTagVisibility and collisionRule to their enum names

u9g:
1.21.5 changed `nameTagVisibility` and `collisionRule` in `packet_teams` from strings to enum ids on the wire. 1.21.6's `proto.yml` maps the ids back to their names, but 1.21.5 was left as a bare `varint`, so it is the only version where these two fields are numbers rather than `always` / `never` / ....

`Team.Visibility` and `Team.CollisionRule` in 1.21.5 use the same `idMapper` and the same order as 1.21.6, so this reuses the 1.21.6 mappers. `protocol.json` regenerated with `tools/js/compileProtocol.js pc 1.21.5`; the diff is only those two fields.

Verified by serializing the same packet through the old and new schema (identical bytes) and parsing it back through the new one (`hide_for_own_team`, `push_other_teams`).

Also adds a `teamModeUsesStringMapper` feature (1.21.6+) for the same packet's `mode` becoming a string mapper, mirroring `entityActionUsesStringMapper`.

Surfaced by PrismarineJS/mineflayer#4081, which needs both to stop special-casing versions in its test.


extremeheat:
CI seems failing for unsorted features.json

Maybe a npm script to auto fix would be helpful

VasilisDragon:
Checked the 1.21.5 `Team.Visibility` and `Team.CollisionRule` codecs in the [client](https://piston-data.mojang.com/v1/objects/b88808bbb3da8d9f453694b5d8f74a3396f1a533/client.jar), using [official mappings](https://piston-data.mojang.com/v1/objects/57669731d542f98646772e91a0d68628f9827a5c/client.txt). Both `STREAM_CODEC`s use `ByteBufCodecs.idMapper`; the IDs and VarInt types match. Numeric rule inputs still work with published ProtoDef 1.19.0’s compiler, but fail with its interpreter and the [upstream compiler](https://github.com/ProtoDef-io/node-protodef/tree/0d7b6e171fe22aaa237a62e7643a5b2ca0dbde2c). Named inputs work in each case.

PR #1303, opened 2026-09-16
pc/protocol: set_cursor_item and set_player_inventory carry a plain Slot on 1.21.3

u9g:
Both packets use `ItemStack.OPTIONAL_STREAM_CODEC` ([ClientboundSetCursorItemPacket](https://github.com/u9g/mc-decompiled/blob/main/src/1.21.3/net/minecraft/network/protocol/game/ClientboundSetCursorItemPacket.java), [ClientboundSetPlayerInventoryPacket](https://github.com/u9g/mc-decompiled/blob/main/src/1.21.3/net/minecraft/network/protocol/game/ClientboundSetPlayerInventoryPacket.java)), whose empty form is a zero item count, not a presence byte. 1.21.4 already declares them as `Slot`; 1.21.3 still had `contents?: Slot`, so an empty cursor parsed as an absent field (`undefined`) and a non-empty one with count 1 would misread the count as the presence flag.

Found via PrismarineJS/mineflayer#4127, where handling `set_cursor_item` crashed on 1.21.3 with `Cannot read properties of undefined (reading 'present')`. Verified against a live 1.21.3 server with the regenerated `protocol.json`.


extremeheat (changes requested):
Are you sure this only applies to 1.21.3? I see multiple versions with similar encoding

[1.21.3](https://github.com/extremeheat/extracted_minecraft_data/blob/client1.21.3/client/net/minecraft/network/protocol/game/ClientboundSetCursorItemPacket.java), [26.3](https://github.com/extremeheat/extracted_minecraft_data/blob/client26.3/client/net/minecraft/network/protocol/game/ClientboundSetCursorItemPacket.java)

rom1504 (commented):

Following up on the version-scope question: I traced the 1.21.x/26.1 protocols selected by `dataPaths.json` at `4cd75dd`. Both packets already use plain `Slot` in 1.21.4, 1.21.5, 1.21.6, 1.21.8, 1.21.9 (also reused by 1.21.10), 1.21.11 and 26.1. Among the schemas present at this head, 1.21.3 is the one needing this correction.

The [1.21.3 ItemStack codec](https://github.com/extremeheat/extracted_minecraft_data/blob/client1.21.3/client/net/minecraft/world/item/ItemStack.java) reads/writes the VarInt item count directly, with zero for empty and no outer presence byte. With NMP 1.68.0's compiled codec, this head correctly encodes/decodes empty, count-1 and count-2 stacks for both packets. CI is green; I found no additional blocker in this change. This was code/codec validation, not a new live-server test.

Skills used: `prismarine-protocol-data-review` for schema routing, version scope and the source/codec comparison; `prismarine-item-inventory-review` for empty/full Slot representation.

PR #1193, opened 2026-05-25
pc: Fix slow break time for harvest-restricted blocks (1.20.5-1.21.11)

VasilisDragon:
## Summary

Blocks with `"material": "incorrect_for_wooden_tool"` and populated `harvestTools` report a `digTime` roughly 9× too slow when broken with a tool that can actually harvest them.

For `diamond_ore` with a netherite pickaxe (item id 943): `materials["incorrect_for_wooden_tool"]` only contains the wooden tool ids (912–915), so the lookup misses, speed defaults to `1.0`, and digTime returns `ceil(1 / (1.0/3.0/30.0)) * 50 = 4550 ms`. Vanilla expects ~500 ms. The block is breakable — `canHarvest` reads `harvestTools`, which is correct — it just takes 9× longer than vanilla.

## Fix

Where `material === "incorrect_for_wooden_tool"` and `harvestTools` is non-empty, switch the material to `"mineable/pickaxe"`. That table is already populated for every pickaxe tier (913, 918, 923, 928, 933, 938, 943 → 2, 1, 4, 12, 6, 8, 9). `harvestTools` is untouched, so `canHarvest` still enforces the tier gate.

`crafter` matches the material but has no `harvestTools`, so it's not touched here.

## Verification

Repro with `prismarine-block` + `minecraft-data`, vs. values from the [Breaking page on the Minecraft Wiki][wiki]:

| block | tool | before | after | vanilla |
|--|--|--|--|--|
| diamond_ore | netherite_pickaxe | 4550 ms | **500 ms** | 500 ms |
| diamond_ore | diamond_pickaxe | 4550 ms | **600 ms** | 600 ms |
| diamond_ore | iron_pickaxe | 4550 ms | **750 ms** | 750 ms |
| iron_ore | iron_pickaxe | 4550 ms | **750 ms** | 750 ms |
| iron_ore | stone_pickaxe | 4550 ms | **1150 ms** | 1150 ms |
| obsidian | netherite_pickaxe | 75000 ms | **8350 ms** | 8350 ms |
| obsidian | diamond_pickaxe | 75000 ms | **9400 ms** | 9400 ms |
| ancient_debris | netherite_pickaxe | 45000 ms | **5000 ms** | 5000 ms |
| netherite_block | netherite_pickaxe | 75000 ms | **8350 ms** | 8350 ms |
| copper_block | iron_pickaxe | 4550 ms | **750 ms** | 750 ms |

[...]

PR #71, opened 2015-09-23
Implement new context syntax

roblabla:
1.9 counterpart of #70

PR #1307, opened 2026-09-18
pc: fix block material for tier-gated blocks (1.20.5 - 26.1)

### LINKED ISSUE #987: Why "incorrect_for_wooden_tool"?
Since version 1.20.5 in data/pc/**/block.json for some blocks in the "material" property it says "incorrect_for_wooden_tool".

I just want to know the reason.
Is it possible to just insert "mineable/pickaxe"? (:

Pix3lPirat3:
Since 1.20.5, blocks with a tool-tier requirement (ores, obsidian, metal blocks, ancient debris; 60–108 per version) have `material: incorrect_for_wooden_tool`. That tag gates drops, not speed, and its `materials.json` table lists only the four wooden tools, so `prismarine-block` `digTime()` misses the lookup for any better tool and falls back to hand speed, also skipping Efficiency. Iron ore with a diamond pickaxe reports 4550 ms instead of 600 ms; with Efficiency V, 30x too slow. Fixes #987; PrismarineJS/mineflayer#3921 and PrismarineJS/mineflayer#4131 are the same bug.

Supersedes #1193 and #1232.

The generator fix is PrismarineJS/minecraft-data-generator#71 (merged July) corrected by PrismarineJS/minecraft-data-generator#86, which also stops #71 from demoting the composite materials that carry sword and shears speeds. No data has been regenerated since PrismarineJS/minecraft-data-generator#71, so this applies that output.

Change: `material` only, `incorrect_for_wooden_tool -> mineable/pickaxe`, for 1.20.5, 1.21.1, 1.21.3, 1.21.4, 1.21.5, 1.21.6, 1.21.8, 1.21.9, 1.21.11 and 26.1. For every directory with a generator module, every block's `material` now equals the PrismarineJS/minecraft-data-generator#86 output. 1.21.4 has no generator module; the same mapping is applied, and all 93 blocks are `mineable/pickaxe` in both the 1.21.3 and 1.21.5 generator outputs. `harvestTools` and `materials.json` are untouched, so `canHarvest` is unchanged.

Verification: `tools/js` suite passes (1860). `digTime()` for every changed block matches the vanilla formula across all tool tiers and types on every version (e.g. 26.1: 3024/3024, was 2160). Live on vanilla 26.1 with mineflayer 4.39.0, iron ore breaks in 607 ms (was 4558 ms) and sword/shears times on melon, leaves and vine are unchanged.

Not included: `drops` (loot sampling differs between generator runs of the same version), and the 1.20.5 `update_1_21` experimental blocks that the generator now resolves with feature-flag packs enabled — a separate cause, left for a 1.20.5 refresh.


Pix3lPirat3:
Updated with the composite fix from generator #86 (second commit there): moss carpet, pink petals and pale moss carpet on 1.20.5 to 1.21.4 are now `plant;mineable/hoe` and bamboo on 1.21.5+ is `sword_instantly_mines;mineable/axe`, with the two merged tables added to `materials.json`, so the hoe and axe speeds are kept next to the sword rules. The 1.20.5 regeneration also corrects 55 blocks the shipped file still had as `default` (tuff variants, chiseled copper, copper doors/trapdoors/grates/bulbs, crafter, heavy core: all `mineable/pickaxe`). Every `material` field in the ten directories now equals the regenerated output

rom1504 (commented):

Looks ready to merge for these material corrections. I checked every changed block field and material-table entry: unrelated block fields, including harvestTools, are preserved; the new composite tables merge their existing component tables correctly. Exact-head generator#86 artifacts agree with all block material fields in 1.20.5, 1.21.3 and 26.1. Production prismarine-block probes across all ten changed directories confirm ore speed and preserve the repaired moss-carpet/hoe and bamboo/axe cases.

Affected schema checks: 97 passing. Current CI is green. The broader local run had 1,895 passing assertions, with failures only in the protocol-sync timeout and overall runtime guard during concurrent execution; I am not reporting that run as a clean full-suite pass. No new live server was run.

Skills used: `prismarine-protocol-data-review` checked generated facts, precise field scope and actual downstream selection; `prismarine-architecture-review` checked the composite representation and producer repair; `prismarine-review` verified the current repaired revision and validation limits.

PR #1080, opened 2025-09-15
Add Minecraft pc 1.21.8 tags

rom1504bot:
This automated PR sets up the relevant boilerplate for Minecraft pc version 1.21.8. Fixes (This issue was created for a minecraft-data-generator PR).

Related:
- Issue: (This issue was created for a minecraft-data-generator PR)
- Protocol Version: null
- https://github.com/PrismarineJS/minecraft-data-generator/pull/56

* You can help contribute to this PR by opening a PR against this <code branch>pc-1_21_8</code> branch instead of <code>master</code>.

Issue #412, opened 2021-06-13
materials.json as a concept is outdated


Archengius:
Material information is used for computing mining speed based on other parameter, but there is one fundamental problem: recent Minecraft versions just do not use materials for anything except colors on the maps.
Instead, mining speed is handled either by the game's code (for swords and shears), or by tags (for all other tools)

Both of these handle blocks individually and independently of their materials, so it's practically impossible to judge mining speeds by block materials anymore. Even more, some blocks can be actually mined up to 3 different tools at once (for example, berry bushes are actually mineable by hoe, axe and sword, yet their material is just "plant").

My suggestion is to get rid of materials.json altogether for newer versions since it's pointless, and record mining speed multipliers directly in blocks.json as values for harvestTools object entries (since practically now information about block breaking speeds is shattered between 2 different files, and now it will be combined into one).
Additionally, boolean indicating whenever block requires tool or not should be added explicitly in this case, instead of relying on harvestTools being absent when block is harvestable by hand.

rom1504:
Do you have a way to retrieve the block speeds in newer versions ?
If yes, could you compare the values you get that way with what
prismarine-block digTime computes ?

Archengius:
Yeah, I have a way to retrieve block speeds in new versions, specifically 1.17.
But I cannot do the comparison because there's simply no minecraft-data for 1.17 to use for prismarine-block.

I could use the data I PR in #411, but comparison would be pointless in that case, since material data I'm PRing is actually based on mining speed categories and not actual materials, to basically have correct mining times without rewriting prismarine-block

Archengius:
Sorry, accidentally closed it. 
Continuing on my last comment, there's also no practical way to retrieve mining speed data on material basis at all in latest version, just because materials are NOT used for it, and you would just get incorrect values for some blocks

rom1504:
the choices for materials are

1. https://github.com/Archengius/minecraft-data-generator/blob/master/src/main/java/me/arch/mcdatagen/generators/MaterialsDataGenerator.java
2. https://github.com/PrismarineJS/burger-extractor + https://github.com/mcdevs/burger
3. https://github.com/PrismarineJS/minecraft-wiki-extractor/blob/master/lib/block_extractor.js#L322 / manually writing the file
choice 2 would be the cleanest, 1 or 3 probably the quickest but not easy to reuse in 1.19
4. import from previous versions https://github.com/PrismarineJS/minecraft-data/pull/282#issue-612896577

[...]

PR #1308, opened 2026-09-18
Add pc tags data extracted from the vanilla server jar (1.13-26.1)

Pix3lPirat3:
Add tools/js/extractPcTags.js, which reads data/minecraft/tags from the official server jar (sha1-verified from the Mojang manifest) and writes data/pc/<version>/tags.json with #tag references resolved, everything sorted and registries namespaced. Generate it for every pc release from 1.13 to 26.1; identical versions share a directory through dataPaths.

Implements the tags direction from #412 and answers the backport question (PrismarineJS/minecraft-data-generator#56, #1080): tags are plain datapack JSON inside every server jar since 1.13, so they need no generator and no mappings. `tools/js/extractPcTags.js` (dependency-free) reads them from the official jar - sha1-verified download from the Mojang manifest, cached in the OS temp dir — resolves `#tag` references, and writes `data/pc/<version>/tags.json`; identical versions share a directory through `dataPaths`, as other data does. `node extractPcTags.js --all` regenerates everything in about two minutes with cached jars.

Data:

27 files covering all 46 pc releases (and pre/rc aliases) from 1.13 to 26.1. Weekly snapshots are excluded because their `blocks.json` aliases unrelated versions. Base datapack only. Registry keys, tag keys and members are namespaced and sorted; empty registries are omitted; every registry in the jar is emitted (block, item, fluid, entity_type from 1.13, growing to twenty on 26.1 including `worldgen/*`).

Format follows #1194's `tags_schema.json`, included verbatim (by @VasilisDragon) with its one-line `test.js` allowlist change so CI validates the data; #1194 remains the home for the audit and schema tests, and I'll rebase onto it if it lands first. Note #1080's file uses bare registry keys in registry order and does not pass that audit; `pc/1.21.8/tags.json` here has identical tag and member sets to it (sorted), so it supersedes #1080.

Verification: all 46 versions pass the schema and #1194's `auditTagDocument` in official mode with block/item/entity_type/biome/enchantment members from each version's own data - no dangling members. `npm test` passes. `1.16` and `1.16.1` alias `pc/1.16-rc1` because the rc precedes the release and is identical; say if a release directory is preferred.

What consumers get: `mineable/*` and `needs_*_tool` from 1.17, `incorrect_for_*_tool` from 1.20.5 - the data `prismarine-block` needs to compute dig time without `materials.json` (#987, #1307). Follow-ups: PrismarineJS/node-minecraft-data#457 to expose `mcData.tags`, then a prismarine-block `digTime()` that reads tags with `materials.json` as the pre-1.17 fallback.


VasilisDragon:
I checked all 46 mapped versions against the official server-jar contents using an independent extractor; the tag sets and members match. `--all` reproduces the committed data, and both `npm test` and the tag audit/schema tests from #1194 pass. The `1.16-rc1` alias also looks fine: its tag data matches both 1.16 and 1.16.1.

PR #1309, opened 2026-09-18
pc: fix copper tool speeds in materials.json (1.21.9-26.1)

Pix3lPirat3:
Copper tools were generated at speed 1.0 because the generator's speed table had no copper entry; vanilla is 5.0. Regenerated with PrismarineJS/minecraft-data-generator#87; 1.21.10 shares pc/1.21.9.

Since 1.21.9 every copper tool is listed at speed 1.0 in `materials.json` - the generator's name-prefix speed table had no copper entry (fixed in PrismarineJS/minecraft-data-generator#87; vanilla `ToolMaterial.COPPER` is 5.0). `prismarine-block` `digTime()` therefore reports copper tools ~5x too slow: copper pickaxe on stone 2300 ms instead of 450, copper shovel on dirt 750 ms instead of 150.

Regenerated `materials.json` for 1.21.9 (also 1.21.10), 1.21.11 and 26.1: exactly 16 entries change per file, all copper tools 1.0 -> 5.0, nothing else. `digTime()` for every diggable block x copper pickaxe/axe/shovel/hoe matches the vanilla formula on all three versions (3887/3888, the remaining case being `bamboo`'s pre-existing missing axe composite, identical for every tier) when combined with #1307. Live on vanilla 26.1 with mineflayer 4.39.0: copper pickaxe on stone 453 ms (was 2307), copper axe on oak log 600 ms (was 3010).


rom1504 (commented):

These copper corrections are already incorporated in master through [minecraft-data#1310](https://github.com/PrismarineJS/minecraft-data/pull/1310), whose merge commit 08a8e3a includes this copper commit plus the shears correction. This open PR therefore appears superseded; it does not need another data merge. Generator#87 (also included in generator#88) remains the producer-side work to preserve the values on regeneration.

I verified all 48 changed entries and compared the 1.21.9/26.1 tables with generator#87's exact-head CI artifacts. Real prismarine-block checks produce the expected copper digging times. Focused schema checks pass (97); the full local run hit timing guards, so I do not claim a clean full-suite run.

Skills used: `prismarine-protocol-data-review` separated landed data from the still-pending generator; `prismarine-architecture-review` checked the shared producer/consumer contract; `prismarine-review` followed actual commit ancestry instead of treating open status as unmerged content.

Pix3lPirat3:
Superseded by #1310 (merged, commit 08a8e3a includes this copper correction plus the shears fix). Generator#87/#88 remain the producer-side work to preserve the values on regeneration.

PR #1315, opened 2026-09-18
pc/1.21.4: recipes from the 1.21.4 generator module

Pix3lPirat3:
The file was the 1.21.3 one: no resin, resin brick or creaking heart recipes, missing dye and suspicious stew variants, and the map-extending recipe as item 0. Regenerated with https://github.com/PrismarineJS/minecraft-data-generator/pull/90: 8 items added, air entry gone, 45 plank recipes re-expanded for pale_oak_planks.

---
**Companion PRs:** generated by PrismarineJS/minecraft-data-generator#90 (1.21.4 module) and PrismarineJS/minecraft-data-generator#89 (skip recipes that assemble to air). Review those alongside; the data here is committed so it does not strictly block.


rom1504 (commented):

No blocking issue found in this data update. I checked all 1,568 candidate entries against the recipes and recursively expanded item tags in Mojang’s official 1.21.4 server jar: each input shape/ingredient multiset and result item/count has a matching vanilla recipe. The previous file’s only unmatched entry in that check was the air result, which this PR removes. The eight new result IDs resolve correctly, the schema validates, and all entries construct through the actual `prismarine-recipe` consumer.

This check covers the represented item IDs, counts and ingredients; it does not claim full component-sensitive crafting or live Mineflayer crafting validation. Existing limitations such as cake remainders are unchanged by this diff.

Skills used: `prismarine-protocol-data-review` checked category routing and generated data against its version-specific source; `prismarine-item-inventory-review` traced the real recipe consumer and separated inherited limitations from this change; `prismarine-review` checked existing feedback.

PR #1318, opened 2026-09-19
pc: language.json from each version's own client, 1.8 - 1.18.2

Pix3lPirat3:
Nine directories served a language file from a neighbouring version: pc/1.16.1 for 1.16.2 - 1.16.5 (63 keys missing, 62 strings different), pc/1.18 for 1.18.2, pc/1.14 for 1.14.3, pc/1.17 for 1.17.1, pc/1.9 and pc/1.10 for their patch releases, and the 1.8 and 1.13.2 files were stale themselves. Files come straight from the client jar's en_us.json (en_US.lang for 1.8); new directories 1.9.4, 1.10.2, 1.14.3, 1.16.2, 1.16.4, 1.17.1, 1.18.2, dataPaths repointed. Every other version already matched its client exactly.


data/pc/1.8/language.json
```diff
@@ -1501,8 +1473,7 @@
   "commands.playsound.success": "Played sound '%s' to %s",
   "commands.playsound.playerTooFar": "Player %s is too far away to hear the sound",
   "commands.give.usage": "/give <player> <item> [amount] [data] [dataTag]",
-  "commands.give.item.notFound": "There is no such item with name %d",
```
rom1504, line 1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please preserve the language data used by 1.8.8/1.8.9 when replacing this shared directory. The Node wrapper resolves those supported versions to the `pc/1.8` dataset (its version record is 1.8.8), so copying the original 1.8.0 client's dictionary here also changes their translations.

I compared the official client jars: the candidate matches 1.8.0 exactly, but differs from 1.8.8 in 40 entries. For example, this deletion removes `commands.give.item.notFound`; `entity.ThrownPotion.name` and several later statistic keys also disappear. Through the real `prismarine-chat` consumer, the potion translation changes from `Potion` to the untranslated key. The old file matches the 1.8.8 dictionary apart from two empty-string entries.

Keep the later patch-version dictionary for its existing consumers, and give original 1.8.0 an explicitly selected dataset if it is to be exposed separately. The other eight changed source dictionaries match their named official clients in my check; this finding is about routing a shared directory, not their extraction.

Skills used: `prismarine-protocol-data-review` traced `toMajor`/category reuse instead of equating a directory name with the selected release; `prismarine-item-inventory-review` checked rendered translation behavior; `prismarine-review` compared the actual base and all existing discussion.

rom1504 (commented):

Found one patch-version routing regression after comparing all nine dictionaries with their official client sources.

Skills: `prismarine-protocol-data-review` checked shared version routing; `prismarine-item-inventory-review` checked the actual chat consumer; `prismarine-review` checked prior feedback.

Pix3lPirat3:
Fixed. Reverted data/pc/1.8/language.json to the 1.8.8 dictionary (pc/1.8's version record is 1.8.8, which the node wrapper resolves 1.8.8/1.8.9 to). The candidate had replaced it with the 1.8.0 client's dictionary, which dropped the ~40 entries 1.8.8 needs (commands.give.item.notFound, the resourcePack.incompatible.* set, tile.pumpkinStem.name, options.entityShadows, entity.ThrownPotion.name, etc.) and changed strings like entity.Arrow.name. The pc/1.8 file now matches master's 1.8.8 data again; the other version language files in this PR are unaffected.

rom1504 (commented):

Verified `df7fe2f`: the shared 1.8 dictionary regression is fixed. The file again matches the previous 1.8.8 data, and real `prismarine-chat` consumers for 1.8.8 and 1.8.9 render `entity.ThrownPotion.name` as `Potion`. The other eight dictionaries still match their named official clients and pass their schemas. This looks ready for maintainer merge review within the checked release scope; I did not independently verify every snapshot alias.

Skills: `prismarine-review` checked the updated head and prior feedback; `prismarine-protocol-data-review` checked version-selected data against native sources and consumer behavior. `prismarine-item-inventory-review` checked the rendered translation rather than just JSON equality.

PR #1319, opened 2026-09-19
pc: fix legacy data pointing at things that do not exist, 1.7 - 1.16.5

Pix3lPirat3:
1.7 used the 1.8 recipe file, so 23 recipes and 5 variants referenced 1.8-only items; it now has its own recipes.json. The 1.8 door recipes were keyed one id too low (426 is not an item), so spruce, birch, jungle, acacia and dark oak doors moved to 427-431. 1.11 items.json lacked iron_nugget (452), which 1.11.2 and the 1.12 snapshots served from that directory craft. 1.13 and 1.13.2 entities.json still carried the pre-flattening object table (28 entries duplicating registry ids 1-94), removed. Wooden tools in 1.13-1.15.2 listed crimson_planks, warped_planks and blackstone as repair materials before those existed. bamboo and bamboo_sapling had material UNKNOWN_MATERIAL in 1.14-1.16.5: bamboo takes the plant table (axe speeds, sword 1.5) that the vanilla axe applies to Material.BAMBOO, the sapling is default like 1.17+.


rom1504 (commented):

Found two introduced version/namespace regressions and one incomplete historical recipe repair. Each finding was checked against the actual data consumer and independent protocol/game evidence.

Skills used: `prismarine-protocol-data-review` traced aliases and ID namespaces; `prismarine-item-inventory-review` checked recipe outputs; `prismarine-architecture-review` checked the shared API; `prismarine-review` checked prior feedback and the baseline.

data/pc/1.13.2/entities.json
```diff
@@ -926,269 +926,5 @@
     "width": null,
     "height": null,
     "type": "mob"
-  },
-  {
-    "id": 1,
-    "internalId": 1,
-    "name": "boat",
```
rom1504, line 933: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please retain the separate object-ID entries in both 1.13 and 1.13.2. Those versions still use a distinct object namespace for `spawn_entity`; the registry IDs do not replace it until the later protocol transition. For example, object ID 1 means boat, whereas registry ID 1 here means armor stand. [ViaVersion's 1.13.2→1.14 translator](https://github.com/ViaVersion/ViaVersion/blob/b4023127f6d1f8d6b0c1f1893e09fe1e2f8a72ac/common/src/main/java/com/viaversion/viaversion/protocols/v1_13_2to1_14/rewriter/EntityPacketRewriter1_14.java#L63-L90) explicitly resolves the byte through `EntityTypes1_13.ObjectType` before converting it to the newer registry ID.

Using the actual Node data loader on the base and candidate files, `objects` goes from 28 entries to zero in both versions; `objects[1]` becomes undefined and `entities[1]` changes from boat to armor stand. These are supported packet/model lookups, not redundant records. Any uniqueness validation needs to distinguish the two namespaces for these versions too.

Skills used: `prismarine-protocol-data-review` distinguished wire object IDs from registry IDs and exercised the production indexer; `prismarine-architecture-review` checked the shared data API; `prismarine-review` checked the exact diff and previous feedback.

data/pc/1.11/items.json
```diff
@@ -2982,6 +2982,12 @@
     "name": "shulker_shell",
     "stackSize": 64
   },
+  {
+    "id": 452,
+    "displayName": "Iron Nugget",
+    "name": "iron_nugget",
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please split the item data at the supported patch-version boundary rather than adding this to the shared 1.11 table. `dataPaths` still routes 1.11 itself, 1.11.2 and the listed 1.12 snapshots to `pc/1.11/items.json`, so the repair now makes `items[452]` / `itemsByName.iron_nugget` exist on 1.11 as well.

I checked SHA1-verified official server jars: 1.11 has no iron-nugget item, while 1.11.2's item-registration bytecode explicitly registers `452, "iron_nugget"`. Keep the 1.11 list unchanged and route 1.11.2 and the applicable later versions to a list containing the nugget. This preserves the missing-item fix without backporting the item into an earlier game version.

Skills used: `prismarine-protocol-data-review` traced all aliases of the edited category and checked official version-specific data; `prismarine-review` distinguished a new regression from the missing 1.11.2 entry being repaired.

data/dataPaths.json
```diff
@@ -14,7 +14,7 @@
       "enchantments": "pc/1.7",
       "effects": "pc/1.7",
       "items": "pc/1.7",
-      "recipes": "pc/1.8",
+      "recipes": "pc/1.7",
```
rom1504, line 17: **Astra agent review — AI-generated, not manually written by the maintainer.**

The dedicated 1.7 file still carries a few 1.8 recipes/quantities, so the historical-version repair is incomplete. It retains slime block output 165, iron trapdoor output 167, and slimeballs 341 made from block 165; neither block exists in the official 1.7.10 registry. It also retains output count 3 for wooden door 324 and iron door 330, whereas the official 1.7.10 crafting-registration bytecode constructs one door for each recipe.

These errors were already present through the old 1.8 alias; I am not calling them new runtime regressions. They are now copied into the new version-specific source introduced to remove exactly this mismatch. Please remove those three recipes and correct both door counts before treating this file as 1.7 data. Do not use the current 1.7 block table alone as the existence oracle: it also contains inherited later-version entries. The [official 1.7.10 server artifact](https://launcher.mojang.com/v1/objects/952438ac4e01b4d115c5fc38f891710c4941df29/server.jar) provides the version-specific registry and crafting evidence used here.

Skills used: `prismarine-protocol-data-review` checked historical source provenance instead of validating one contaminated table against another; `prismarine-item-inventory-review` checked craft result quantities and actual recipe-model loading; `prismarine-review` separated inherited errors from new regressions.

Pix3lPirat3:
All three findings addressed in 14865f8c:

1. **1.13/1.13.2 object IDs**: restored the 28 `type: object` entries in both entities.json (reverted to master). spawn_entity keeps its separate object namespace through 1.13.2 (object ID 1 = boat, distinct from registry ID 1 = armor stand), so those entries must stay until the 1.13.2->1.14 transition.

2. **1.11 iron_nugget**: reverted pc/1.11/items.json (no iron_nugget), and split the item data at the patch boundary - new pc/1.11.2/items.json carries iron_nugget (id 452), and dataPaths routes 1.11.2, 17w15a, 17w18b and 1.12-pre4 to it. 1.11 stays without it; 1.11.2+ get it. Verified: 1.11 iron_nugget=false, 1.11.2/17w15a=true.

3. **1.7 recipes**: removed the 1.8-only crafts (slime_block 165, iron_trapdoor 167, and the 9-slimeball-from-slime_block recipe 341), and set wooden_door (324) and iron_door (330) output count 3->1 to match 1.7.10's one-door-per-recipe.

Cross-file consistency test (with the #1320 fixes): 674 passing, 107 pending, 0 unexpected failures.

Pix3lPirat3:
(cross-ref) The loot producer fix for #1312 landed in PrismarineJS/prismarine-loottable#15 (additive set_count handling).

rom1504 (commented):

Verified `14865f8`: all three findings from my earlier review are fixed. The production Node loader again exposes 28 objects in both 1.13 and 1.13.2 (`objects[1]` is boat, while `mobs[1]` is armor stand). The item-category split keeps iron nuggets absent on 1.11 and present on 1.11.2 and the three routed snapshots. The dedicated 1.7 recipe file removes the three later crafts and gives both door recipes one output, matching the previously checked official game evidence.

The 12 changed data files pass their schemas. I also rechecked the retained door-key, repair-material and bamboo changes against the full diff. I have no remaining blocker from this review; this looks ready for maintainer merge review.

Skills: `prismarine-review` checked the updated head and prior feedback; `prismarine-protocol-data-review` checked version-selected data against native sources and consumer behavior. `prismarine-item-inventory-review` checked historical recipe outputs and the supported patch boundary.

PR #1320, opened 2026-09-19
tools/js: test cross-file consistency of the pc data

Pix3lPirat3:
Every version's ids and names must be unique, registry ids (items, entities, sounds, and effects, enchantments and biomes from 1.20.3) must start at 0 without gaps, block state ids must be contiguous with the default state in range, and everything that points at something else must point at something that exists: block drops and harvest tools, material speed tables, foods, repairWith, recipe ids, block and entity loot names, enchantment exclusions and costs, collision shape coverage. The current data fails 152 of the 781 checks; those are listed in pc_consistency_known.json with the fix that is in flight for each and run as pending, and the run prints which entries can be dropped once they pass, so a fix removes its own entries and any new inconsistency fails the build.


tools/js/test/pc_consistency.js
```diff
@@ -0,0 +1,181 @@
...
+const TOOL = /(_pickaxe|_axe|_shovel|_hoe|_sword|^shears$|^spade$)/
+const isTool = (name) => TOOL.test(name)
+const list = (arr, n = 8) => arr.slice(0, n).join(', ') + (arr.length > n ? `, ... (${arr.length} total)` : '')
+
+for (const version of versions) {
+  if (!dataPaths[version]) continue
+  describe(`pc ${version} consistency`, function () {
+    const check = (title, fn) => it(title, function () {
+      const key = `${version} | ${title}`
+      try { fn() } catch (err) { if (known[key]) return this.skip(); throw err }
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please narrow the known-failure exception to the recorded offending entries, rather than skipping any exception from this whole version/check. The entry for `1.8 | ids and names are unique` describes an existing duplicate biome, but it also hides a newly duplicated **item**. I appended another `stone` item to 1.8/items.json: this test still exits successfully as pending. Removing only the known-biome exception makes the same mutation fail with `items: duplicate ids stone#1`.

The problem also persists after a fix lands until its known key is manually removed: an unrelated new regression under that key silently becomes pending again. Record/compare the expected offending category and IDs/names (or otherwise fingerprint specific failures), fail on additional problems, and do not swallow arbitrary non-assertion errors. This preserves a usable baseline while making the promised new-regression protection real.

Skills used: `prismarine-protocol-data-review` checked whether the generic validator catches the defect class; `prismarine-review` required a mutation/negative control instead of accepting a green baseline as evidence.

tools/js/test/pc_consistency.js
```diff
@@ -0,0 +1,181 @@
...
+    const flattened = atLeast(version, '1.13')
+    // before the flattening, block ids below 256 double as item ids
+    const isItemId = (id) => itemsById.has(id) || (!flattened && blocks.some(b => b.id === id && id < 256))
+
+    check('ids and names are unique', function () {
+      for (const [kind, data] of [['blocks', blocks], ['items', items], ['entities', load(version, 'entities')], ['biomes', load(version, 'biomes')], ['effects', load(version, 'effects')], ['enchantments', load(version, 'enchantments')], ['sounds', load(version, 'sounds')]]) {
+        if (!Array.isArray(data)) continue
+        const dupIds = []; const dupNames = []; const ids = new Set(); const names = new Set()
+        for (const e of data) {
+          const idKey = kind === 'entities' && !flattened ? `${e.type}/${e.id}` : e.id // legacy mobs and objects have separate id spaces
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

The separate entity/object ID spaces continue through **1.13.2**, independently of block flattening. In those versions `spawn_entity` still carries an object-type byte: object ID 1 is a boat, while registry/mob ID 1 is an armor stand. [ViaVersion's actual 1.13.2 → 1.14 translator](https://github.com/ViaVersion/ViaVersion/blob/b4023127f6d1f8d6b0c1f1893e09fe1e2f8a72ac/common/src/main/java/com/viaversion/viaversion/protocols/v1_13_2to1_14/rewriter/EntityPacketRewriter1_14.java#L63-L90) explicitly converts that ObjectType into the later registry ID.

Consequently the 1.13/1.13.2 duplicates recorded as known failures here are valid separate namespaces; deleting the object table in #1319 would break `objectsById`. Keep the type-qualified uniqueness/name rules through this transition, and apply the contiguous-registry check below to the registry table rather than the sparse object-ID table. Use the entity protocol transition instead of reusing the block `flattened` condition.

Skills used: `prismarine-protocol-data-review` distinguished object wire IDs from registry IDs and checked the real version boundary; `prismarine-architecture-review` traced the public `objectsById` consumer; `prismarine-review` coordinated this with the companion data correction without endorsing deletion of valid data.

rom1504 (commented):

The baseline suite runs as reported: 629 passing and 152 pending. Two issues need correction before relying on it as a regression gate: known exceptions currently suppress unrelated new defects, and the entity namespace checks treat block flattening as the object-ID transition. The first is reproduced by an independent duplicate-item mutation; the second is checked against the actual 1.13.2→1.14 protocol translation and public object lookup. Details are inline.

Skills used: `prismarine-protocol-data-review` checked validator scope and ID namespaces; `prismarine-architecture-review` traced the loader API; `prismarine-review` verified current discussion and negative controls.

Pix3lPirat3:
Both points addressed in 6da80d2d.

1. Known-failure narrowing: pc_consistency_known.json is now keyed to { reason, error } and a known failure is tolerated only when the actual assertion message still matches `error` exactly - a new or changed failure of the same check throws. Confirmed with your mutation: appending a duplicate `stone` to 1.8/items.json now fails with `items: duplicate ids stone#1` instead of passing as pending. Added a RECORD_KNOWN=1 mode to regenerate the file (keeps each reason, records the current error).

2. Entity/object id space: introduced `legacyEntityIds = !atLeast(version, '1.14')` and use it (not `flattened`) for the entity id-key, the class-name-reuse skip, and the contiguity check - so the separate object id space is honored through 1.13.2. The 10 now-obsolete 1.13/1.13.1/1.13.2(-pre) entity known-failures were dropped.

Full run: 639 passing, 142 pending, 0 unexpected failures.

tools/js/test/pc_consistency.js
```diff
@@ -0,0 +1,207 @@
...
+  describe(`pc ${version} consistency`, function () {
+    const check = (title, fn) => it(title, function () {
+      const key = `${version} | ${title}`
+      const entry = known[key]
+      try {
+        fn()
+      } catch (err) {
+        if (recording) { recorded[key] = { reason: (entry && entry.reason) || (typeof entry === 'string' ? entry : 'unclassified'), error: err.message }; return this.skip() }
+        // Skip only the exact recorded failure; a new or changed failure of the same check must still throw.
+        if (entry && entry.error === err.message) return this.skip()
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

The earlier duplicate-stone probe now fails correctly, but matching the **first** assertion message still hides additional failures later in this check. With the current 1.8 data, I appended a duplicate `Speed#1` effect: the test still exits successfully with `0 passing, 1 pending`, because the known duplicate biome throws before the effects category is examined. Removing only that existing duplicate biome makes the same mutation fail with `effects: duplicate ids Speed#1`.

Please gather the complete set of problems across the check before comparing it with the known baseline (or separate the categories/assertions). Compare complete, untruncated problem identities; the display message should not be the fingerprint. This is the remaining part of the earlier exception concern, not a request to repeat the now-fixed stone case.

Skills: `prismarine-review` verified the repair against prior feedback; `prismarine-protocol-data-review` checked the complete regression gate with mutation controls; `prismarine-architecture-review` checked the shared validator contract.

rom1504 (commented):

Verified `6da80d2`: the 1.13/1.13.2 object-namespace correction is fixed. The suite runs with 639 passing and 142 pending. The known-failure change catches the previous stone mutation, but another independent mutation shows that a known earlier assertion still prevents later categories from being checked; details inline. That leaves this regression-gate repair incomplete.

Skills: `prismarine-review` verified the repair against prior feedback; `prismarine-protocol-data-review` checked the complete regression gate with mutation controls; `prismarine-architecture-review` checked the shared validator contract.

PR #1321, opened 2026-09-19
bedrock: add 1.26.51 (protocol 2193)

Pix3lPirat3:
The 1.26.50 protocol as documented by gophertunnel 520: play_sound gains bypass_listener_range_check and an optional playback_position_seconds, move_entity_delta gains ticks, boss_event loses player_id, disconnect gains missing_structure_data and unsupported_transport, the outer presence bools in player_auth_input, inventory actions and item stack responses are gone (single optionals now), inventory transactions always carry actions and the use-item data carries the hand, dimension definitions send min_y, height_range and default_biome, pack settings can be a string list, sub chunk height maps are 16 length-prefixed rows, entity diagnostics carry position and dimension, and two packets are new: set_player_furnace_options (0x15f) and record_started (0x160). Camera presets also gain apply_inherited_starting_rotation and starting_rotation. The 1.26.45 directory keeps its own proto and types snapshot. Verified by logging into an official Bedrock Dedicated Server 1.26.51.1 over NetherNet with bedrock-protocol: login, encryption, start_game, item registry, biomes, camera presets, chunks and spawn all decode, a real 1.26.51 client relayed through the same definitions produced no decode failures or packet violations across 25,000 packets, and sample dumps built from the server pass bedrock-protocol's internal client and server replay test.


data/dataPaths.json
```diff
@@ -2921,6 +2921,28 @@
       "types": "bedrock/latest",
```
mc-zuri, line 2921: update proto and types paths to 1.26.40 version.

data/bedrock/latest/types.yml
mc-zuri, line 3477: `ShapeText` is missing `line_gap_height: lf32`**

`TextDataPayload` gains `LineGapHeight` at ordinal 3, between `BackgroundColor` and `DepthTest`:
[v1.26.51 `TextDataPayload.json#L22-L28`](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/TextDataPayload.json#L22-L28) (not in [v1.26.45](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.45/json/TextDataPayload.json)).

On the wire it is a plain float, not an option: a 1.26.51 server's own reader accepts
`08 "Welcome!" 00 00 0000c03f 01 00 00` — `use_rotation`, absent colour, `1.5f`, then the three bools.

Claude

mc-zuri, line 3398: `EnvironmentAttributeData` is missing `noise_alignment`**

A required `NoiseAlignment` is appended at ordinal 9, after `NoiseTransition`:
[v1.26.51 `EnvironmentAttributeData.json#L62-L65`](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/EnvironmentAttributeData.json#L62-L65), and it is in the [`required` list](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/EnvironmentAttributeData.json#L99).

Both of its types are new in this version: [`NoiseAlignment.json`](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/NoiseAlignment.json) (`type: u8` enum, then `value: uint32` compressed) and [`NoiseAlignmentType.json`](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/NoiseAlignmentType.json) (`minLocalTransitionEnd`).

[`EnvironmentAttributeData`](https://github.com/PrismarineJS/minecraft-data/blob/adc702ff2935dabc1a23f61324898950b2af512c/data/bedrock/latest/types.yml#L3389-L3398) ends at `noise_transition`. It is an array element in `packet_clientbound_attribute_layer_sync`, so every non-empty packet comes up short.

Claude

mc-zuri: `GatheringJoinInfo` — did the remaining required fields become optional?** (question, not certain)

In v1.26.45 all 8 fields are in the [`required` list](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.45/json/gatheringsConfig.json#L64-L73); in [v1.26.51](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/json/gatheringsConfig.json) that list is gone. Mojang's note for this release says the same: [`r26u5/us_1667208.md`](https://github.com/Mojang/bedrock-protocol-docs/blob/v1.26.51/developer_notes/r26u5/us_1667208.md) — "making packet fields optional".

[`GatheringJoinInfo`](https://github.com/PrismarineJS/minecraft-data/blob/adc702ff2935dabc1a23f61324898950b2af512c/data/bedrock/latest/types.yml#L2672-L2680) still has `experience_id`, `experience_name` and `creator_id` as required. It is used by `packet_transfer` and by `ServerJoinInformation` in `start_game`. Only gatherings-enabled servers would hit it, so I couldn't confirm against real traffic.

data/dataPaths.json
mc-zuri, line 2899: update types path to 1.26.40 version.

Pix3lPirat3:
Thanks for the thorough review - these are exactly the 1.26.51 wire deltas I wanted a second set of eyes on.

Status on our side: our local `bedrock/1.26.51` is already fast-forwarded onto your `bedrock/protocol` tip, and it's been verified end to end - offline `test/bedrock/*` (67) plus a live soak against an official BDS 1.26.45 (craft/inventory/equip/chest/movement/smelt, 0 kicks/0 errors). So your 55 fixes are running cleanly through the adapter.

Plan per finding:
- **ShapeText `line_gap_height`** - will add `line_gap_height: lf32` (required, per your "plain float, not an option") at ordinal 3, between `background_color` and `depth_test`. I'll round-trip your `08 "Welcome!" 00 00 0000c03f 01 00 00` bytes to confirm the trailing fields line up.
- **EnvironmentAttributeData `noise_alignment`** - will append the required `noise_alignment` at ordinal 9 and add `NoiseAlignment` (`type` u8 enum + the compressed `value`) and `NoiseAlignmentType` (`minLocalTransitionEnd`). One check for you: what's the ProtoDef type for the "uint32 compressed" value - `varint`, or a specific compressed type you're already using elsewhere?
- **GatheringJoinInfo** - I'll confirm against a real gatherings packet (or your harness) before flipping the three fields to optional; agree the Mojang note points that way.
- **dataPaths (proto/types)** - will freeze 1.26.40/1.26.45 to their own `proto`/`types` so the 1.26.51-only fields don't leak back into the older versions via `bedrock/latest`.

One coordination question on structure: since our branch already sits on top of yours, would you prefer these 1.26.51 fixes land in your `bedrock/protocol` branch (and #1321 rebases to just add 1.26.51 on top of it / or closes in favor of it), or that #1321 carries the whole set? Happy either way - I can run your `bedrock-protocol-tests` round-trips on whichever branch to verify before it merges.

mc-zuri:
looks like GatheringJoinInfo change log message was misleading. there is no change in 1.26.51.1 or it does not affected protocol format. 

feel free to cherry pick any commit from that branch if you need. I will do review changes and create new smaller branches for pull requests when your pull request will be merged.

Pix3lPirat3:
Thanks - dropping GatheringJoinInfo then (no real 1.26.51.1 change, agreed).

On structure: since you'll spin your other fixes into their own smaller PRs after this merges, I'll keep #1321 **focused on 1.26.51** - the version bump plus only the two 1.26.51-specific wire deltas you found:
- ShapeText: `line_gap_height: lf32` at ordinal 3 (I'll round-trip your `08 "Welcome!" 00 00 0000c03f 01 00 00` to confirm).
- EnvironmentAttributeData: append `noise_alignment` at ordinal 9 + the `NoiseAlignment` / `NoiseAlignmentType` types.

That keeps the diff reviewable and leaves your cross-version fixes (1.16-1.26.45) to your PRs. If you'd rather this PR carry the whole set instead, say the word and I'll push the fast-forward.

One thing still open: for `NoiseAlignment`'s "compressed uint32" value - is that a plain `varint`, or a specific compressed type you're already using elsewhere in the bedrock schema? Want to match exactly what BDS emits. I'll verify whatever we land with your bedrock-protocol-tests round-trips before pushing.

extremeheat:
@Pix3lPirat3 following the maintainer's updated plan, I pushed `d0d6f19` to this PR using the enabled maintainer-edit access. The Bedrock CI checkout now explicitly uses `PrismarineJS/bedrock-protocol`'s `nethernet` branch, so we can test the data and transport together without merging bedrock-protocol#774 into master first.

That branch now uses published `nethernet@^1.1.0`. The workflow override is temporary: restore the default/master checkout once #774 lands.

This only changes which Bedrock revision CI tests. The existing vanilla spawn/packet-dump harness still selects RakNet; if the 1.26.51 connection timeout persists, the next step is to reproduce the server's transport setup and adapt those test clients accordingly, rather than infer a schema failure from the timeout. The exact data commit and vanilla-server setup from your successful run would still be helpful.

— OpenAI Astra

extremeheat:
Found and reproduced the timeout locally. The BDS 1.26.51.1 distribution has `transport=nethernet` in server.properties, but our vanilla spawn and packet-dump clients use RakNet. Both sides were waiting on different transports; increasing the timeout would not fix that.

Pushed the small test-harness fix to bedrock-protocol's `nethernet` branch at `c4d5aca`: `tools/startVanillaServer.js` now defaults the launched server to `transport=raknet`, matching those clients. Explicit callers can still override the transport. This avoids adding Nethernet discovery/signalling to the packet-schema harness just to test the data.

Validation using this PR's data (`d0d6f19`):
- Before: reproduced `client spawns 1.26.51: Connect timed out` against official BDS 1.26.51.1.
- After: both vanilla clients spawned, packet dumping completed, and the internal replay test passed for 1.26.51 (2 tests).
- The same spawn/dump/replay checks passed for 1.26.45 (2 tests).
- Lint, types and all 90 focused tests passed with this PR's data.

BDS prints a warning that Nethernet is the supported transport, but its RakNet compatibility path completed these tests. This fixes the existing schema harness; it does not add online or vanilla Nethernet coverage.

Reran the failed Bedrock CI job; it checks out the updated `nethernet` branch. No merge into Bedrock master or npm release was made.

— OpenAI Astra

extremeheat:
Added `defaultTransportIsNethernet` to Bedrock's common features at the maintainer's request. Consumers can query `mcData.supportFeature('defaultTransportIsNethernet')` instead of hard-coding the transport's version boundary.

The default changed in 1.26.50, but this PR's version table jumps from 1.26.45 to 1.26.51. The feature resolver requires a mapped lower bound, so the range starts at 1.26.51 (the first mapped affected release) through latest, with the historical boundary documented in the description. Checked that the flag is false for 1.16.201/1.26.40/1.26.45 and true for 1.26.51. Once 1.26.50 is mapped, move the lower bound to that version.

This records a default, not exclusive support: explicitly configured transport should win. The existing vanilla schema harness still intentionally configures RakNet. Testing the new default over Nethernet also needs the vanilla server's connection/discovery/signalling path; changing just the feature flag or client transport option is not enough.

— OpenAI Astra

extremeheat:
The feature-driven transport tests are now green. This supersedes the temporary RakNet-only workaround above.

Bedrock's `nethernet` branch at `00c742e` reads `defaultTransportIsNethernet` to select the transport for both the vanilla server and its spawn/packet-dump clients. Older versions use RakNet; 1.26.51 uses Nethernet through the dedicated server's HTTP signalling endpoint. The local offline test helper gathers ICE candidates and includes the offline identity assertion before sending the offer. Rejected offers and cancellation have regression coverage.

[CI attempt 2 passed with 149 tests](https://github.com/PrismarineJS/minecraft-data/actions/runs/35564235483/attempts/2). Its log confirms checkout of Bedrock `00c742e` and this PR's `19d182f` merge ref, including successful 1.26.51 spawn, packet dumping and replay. [Bedrock's own Linux and Windows CI also passed](https://github.com/PrismarineJS/bedrock-protocol/actions/runs/35565486318).

The workflow can continue targeting `nethernet` while these PRs are reviewed; restore its default checkout after bedrock-protocol#774 lands. No merge into Bedrock master or release was needed. This validates local offline BDS and packet schemas; live Realm/friend-world testing remains separate.

— OpenAI Astra

PR #1325, opened 2026-09-22
Add shared Nethernet advertisement schemas

extremeheat:
Move the Nethernet advertisement ProtoDef definitions from bedrock-protocol into shared Minecraft data. The file contains the v4/v7 layouts and the optional v4 trailer. Discovery layouts are independent of the advertised game version, so this belongs in `data/bedrock/common` and can be decoded before a Minecraft version is known.

This is an unchanged extraction of `src/nethernet/advertisement.json` from PrismarineJS/bedrock-protocol. Bedrock will consume it through the minecraft-data npm package, allowing minecraft-bedrock-server to use Bedrock's shared discovery API instead of carrying duplicate schemas and transport code.


extremeheat:
Closing the schema extraction: the advertisement definitions will remain in bedrock-protocol. extremeheat/minecraft-bedrock-server#37 can use its shared discovery API directly, so this move is unnecessary.

